PR #2, opened 2015-11-11
Everything should compile on the first attempt.

tomas-vondracek:
Created maven parent to hold the java module as well as the server module. With this structure, the whole project can be compiled and packaged with one maven command. It can be imported into the IDE easily as well.

If you intended to use different kind of building, feel free to discard this pull request.


petr-inmite:
Thank you for the pull request! :)

PR #45, opened 2016-02-01
Single crypto provider module instead of 3.

tomas-vondracek:
Merged powerauth-java-prov, powerauth-java-prov-spongycastle, powerauth-java-prov-bouncycastle modules into single one. The decision what security provider to use is made at runtime based on whichever one is registered.


petr-inmite:
Does that mean that we package both (and all possible) provider JARs into the resulting WAR / JAR despite the fact only one of them is used?

tomas-vondracek:
Nope. In the powerauth-java-prov module, the *castle dependencies are scoped as "provided". That means the dependency is used to compile the module, but it's not added to the classpath. Therefore the powerauth-java-server, powerauth-android, powerauth-java-cmd (the runnable artifacts) need to explicitly declare either bouncycastle or spongycastle dependency. There would be no castle at runtime otherwise.

tomas-vondracek:
You can build the WAR/JAR and have a look inside :)

petr-inmite:
Thank you! :)

PR #95, opened 2016-08-16
Added TrustManager 

ForestTree:
Added TrustManager that allows connection to servers with unsecured self-signed certificates


petrdvorak:
Thank you for the pull request!

PR #146, opened 2017-11-15
Fix #145: Prepare PA token computation routine 

### LINKED ISSUE #145: Prepare PA token computation routine

petrdvorak:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/TokenUtils.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            return String.valueOf(System.currentTimeMillis()).getBytes("UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            // ... in case system does not support UTF-8
+        }
+        return null;
+    }
+
+    /**
+     * Helper method to convert provided timestamp into 8 bytes, for the purpose of token timestamping.
```
hvge: This is kind of misleading. It's actually 13 characters (or bytes when converted to utf8) for current epoch.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/TokenUtils.java
```diff
@@ -0,0 +1,116 @@
...
+        } catch (UnsupportedEncodingException e) {
+            // ... in case system does not support UTF-8
+        }
+        return null;
+    }
+
+    /**
+     * Compute the digest of provided token information using given token secret.
+     * @param nonce Token nonce, 16 random bytes.
+     * @param timestamp Token timestamp, Unix timestamp format encoded as 8 bytes.
```
hvge: The same 8-byte mistake as above.

PR #165, opened 2018-06-08
Pull request for 0.19.0 release

petrdvorak:
This is an open pull request for the purpose of evaluating changes made for 0.19.0 release.


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/KeyGenerator.java
```diff
@@ -178,13 +178,26 @@ public SecretKey deriveSecretKeyHmac(SecretKey secret, byte[] index) {
...
+    /**
+     * Derive a long AES suitable key from a password and salt. Uses PBKDF with specified
+     * number of iterations.
+     *
+     * @param password A password used for key derivation.
+     * @param salt A salt used for key derivation.
+     * @param iterations Number of iterations used in PBKDF.
+     * @return A new secret key derived from the password.
+     */
+    public SecretKey deriveSecretKeyFromPassword(String password, byte[] salt, int iterations) {
```
petrdvorak: While we no longer need this method, I suggest keeping it here - having ability to set iteration count to PBKDF2 seems helpful.

PR #190, opened 2018-09-12
Fix #185: Crypto 3.0: Hash based counter

### LINKED ISSUE #185: Crypto 3.0: Hash based counter
Implement new hash based counter as defined in specification:

Let's call this new hash-based counter as `CTR_DATA`, which must be 16 bytes long. I'll use following terms in documentation:

- `CTR` - current linear counter; 64 bit, big endian ordered integer.
- `CTR_DATA` - new non-incremental counter; array of 16 bytes.
- `CTR_DATA_next` - a next value for `CTR_DATA`

Proposed changes in counter related tasks:

- **New activation:**
  Server generates first CTR_DATA and sends it to the client during the initial keys exchange:
  ```
  CTR_DATA = Generator.randomBytes(16)
  ```

- **Calculating signature**
  - In the signature calculation routine, we're just use `CTR_DATA` as it is, instead of original `CTR`.
  - Server must decide which value must be used for calculation, depending on the version of the protocol.
  

- **Calculating next value** (equivalent to `CTR = CTR + 1`)
  ```
  CTR_DATA_next = ByteUtils.convert32Bto16B(Hash.sha256(CTR_DATA))
  ``` 
 
- **Migrating old activation** requires two new endpoints (see separate chapter)

romanstrobl:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/HashBasedCounterGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.getlime.security.powerauth.crypto.lib.util.Hash;
+
+/**
+ * Generator of hash based counter.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ *
+ */
+public class HashBasedCounterGenerator {
```
petrdvorak: I would consider simplifying the naming here a bit:

```java
HashBasedCounter ctr = new HashBasedCounter(); // implements Counter...
byte[] seed = ctr.init();
byte[] next = ctr.next(seed);
```

The idea is that this is a very simple internal class and we would like to emulate this code as simple as possible:

```java
int i = 0;
i++;
```

romanstrobl: Yes, it's more concise, let's use it.

petrdvorak:
Looks OK for me.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/server/vault/PowerAuthServerVault.java
```diff
@@ -41,9 +41,12 @@
...
+     *
+     * PowerAuth protocol version: 2.0
+     *
      * @param serverPrivateKey Server private key KEY_SERVER_PRIVATE
      * @param devicePublicKey Device public key KEY_DEVICE_PUBLIC
-     * @param ctr Counter CTR
+     * @param ctr Numeric counter.
      * @return Encrypted vault encryption key.
      * @throws InvalidKeyException In case a provided key is incorrect.
      */
```
hvge: This also may be marked as deprecated.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/vault/PowerAuthClientVault.java
```diff
@@ -40,9 +40,12 @@
...
+     *
+     * PowerAuth protocol version: 2.0
+     *
      * @param cVaultEncryptionKey Encrypted vault encryption key KEY_ENCRYPTION_VAULT.
      * @param masterTransportKey Master transport key used for deriving a transport key, used for decrypting vault encryption key.
-     * @param ctr Counter used for key derivation.
+     * @param ctr Numeric counter used for key derivation.
      * @return Original KEY_ENCRYPTION_VAULT
      * @throws InvalidKeyException In case invalid key is provided.
      */
```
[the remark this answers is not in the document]
romanstrobl: Ok

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/server/vault/PowerAuthServerVault.java
romanstrobl: Yes

romanstrobl:
All issues are resolved, I am merging the pull request.

PR #197, opened 2018-09-18
Fix #195: Crypto 3.0: New test vectors for ECIES

### LINKED ISSUE #195: Crypto 3.0: New test vectors for ECIES
Add test which uses new test vectors for ECIES shared with the mobile SDK.

romanstrobl:
Note that we may still refactor ECIES a bit for Crypto 3.0 (e.g. change deriving of `sharedInfo1` which is now present in the test, consider using the `ECIESEnvelopeKey` like in the mobile SDK, etc.). So at least we can easily test such changes because of `v2` and `v3` tests.

### DIFF
--- a/powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/encryption/BasicEciesEncryptorTest.java
+++ b/powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/encryption/BasicEciesEncryptorTest.java
@@ -267,155 +267,12 @@ public class BasicEciesEncryptorTest {
     }
 
     /**
-     * Test that values match client side test vectors.
-     *
-     * <h5>PowerAuth protocol versions:</h5>
-     * <ul>
-     *     <li>2.0</li>
-     *     <li>2.1</li>
-     * </ul>
-     *
-     * @throws Exception When test fails.
-     */
-    @Test
-    public void testVectorsV2() throws Exception {
-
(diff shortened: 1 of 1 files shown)


powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/encryption/BasicEciesEncryptorTest.java
```diff
@@ -268,10 +268,17 @@ public void testKdf() {
...
+     * <ul>
+     *     <li>2.0</li>
+     *     <li>2.1</li>
+     * </ul>
+     *
      * @throws Exception When test fails.
      */
     @Test
-    public void testVectors() throws Exception {
+    public void testVectorsV2() throws Exception {
```
hvge: I think that we don't need to split this into v2 & v3. At the fundamental level, v3 is simply superset of v2. If the sharedInfo1 parameter is empty, then the computation is equal to v2. The new vectors already covers both situations.

You can use that old data if you want, there's nothing wrong with that, but we don't need to highlight that it's a different version. We will not have two different ECIES implementations in the code. Just one, with right parameters for both, v2 & v3 protocols.

romanstrobl: Ok, I will clean it up.

romanstrobl:
Alright, I removed the old test vectors.

PR #246, opened 2019-02-28
Update documentation link

romanstrobl:


romanstrobl:
As agreed, I updated the link to contains the `current` version instead of `latest`.

PR #253, opened 2019-05-09
Fix #252: Detect duplicate PUKs when generating recovery codes

### LINKED ISSUE #252: Detect duplicate PUKs when generating recovery codes
In current implementation the PUK derivation indexes are checked for uniqueness which does not guarantee unique PUKs. We should check for unique PUKs instead.

romanstrobl:
Detection of duplicates is now done on `PUK`s, not on `pukDerivationIndex`es.

### DIFF
--- a/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
+++ b/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
@@ -224,7 +224,7 @@ public class IdentifierGenerator {
             Long derivationIndex;
             String derivedPuk;
             do {
-                // Generate random derivation index which must be unique
+                // Generate random derivation index
                 derivationIndexBytes = keyGenerator.generateRandomBytes(8);
                 derivationIndex = ByteBuffer.wrap(derivationIndexBytes).getLong();
                 // Generate recovery PUK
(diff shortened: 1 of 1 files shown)


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
```diff
@@ -220,17 +220,19 @@ public RecoveryInfo generateRecoveryCode(SecretKey secretKey, int pukCount, bool
...
-                // Generate random derivation index which must be unique
-                derivationIndexBytes = keyGenerator.generateRandomBytes(8);
+                // Generate random derivation index
+                byte[] derivationIndexBytes = keyGenerator.generateRandomBytes(8);
                 derivationIndex = ByteBuffer.wrap(derivationIndexBytes).getLong();
-            } while (pukDerivationIndexes.values().contains(derivationIndex));
+                // Generate recovery PUK
+                derivedPuk = generatePuk(recoveryPukBaseKey, derivationIndexBytes);
+                // Make sure that generated PUK is unique
+            } while (puks.values().contains(derivedPuk));
```
petrdvorak: There should be some cycle termination condition for when the universe is broken.

romanstrobl: Ok, I will add it.

PR #269, opened 2019-06-19
Fix classloader issues when using Bouncy Castle library

romanstrobl:
I would like to integrate this pull request early for the next release so that we have enough time to review and test the changes.

This pull request resolves the infamous "key spec not recognized" error when multiple sibling classloaders are used and both classloaders load the Bouncy Castle library. The issue is easily reproducible on Tomcat when two war files contain the Bouncy Castle library. BC library compares instances of `KeySpec` which differ when loaded by two classloaders when the classloader are not in a parent-child (or a parent-ascendant) relationship. Hence the key specification instance is not recognized and defaults from EC `KeyFactorySpi` to `BaseKeyFactorySpi` and returns the infamous error.

The fix migrates usage of Bouncy Castle API for key conversion to standard Java Crypto API and still using Bouncy Castle for the actual conversion from/to bytes. This way the comparison of key specification in Bouncy Castle library is avoided because standard Java Crypto API is used instead. The conversion operations using Bouncy Castle library are safe because they only deal with regular `BigInteger` and `byte[]` which does not cause anly classloader issues.

The Java Crypto API still uses Bouncy Castle as an implementation, so all EC point validity checks remain as before and generally the behaviour should be identical except for initialization of crypto objects which is routed through Java Crypto API instead of BC directly.

I can look into writing some tests to prove that the new implementation is equivalent to the previous implementation (perhaps next month), however this should be quite evident from using the code to convert private and public keys. So I suggest we start testing this early and cover the fix with tests a bit later on, too.

### DIFF
--- a/powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/activation/PowerAuthActivationTest.java
+++ b/powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/activation/PowerAuthActivationTest.java
@@ -34,6 +34,7 @@ import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.Security;
+import java.security.interfaces.ECPublicKey;
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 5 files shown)


romanstrobl:
Ok, I am really merging this change, let's see what happens next :)

PR #274, opened 2019-06-27
Update coverity scan settings

romanstrobl:
Update of coverity scan configuration

I am moving coverity scan to a separate branch (as recommended by Coverity Scan). The service has frequent outages (e.g. last upgrade caused a 4-day outage, earlier the service was inaccessible for weeks), we don't want to show failing builds whenever coverity scan fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ jdk:
   - oraclejdk8
 env:
   global:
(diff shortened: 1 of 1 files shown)


saalistaja:
@romanstrobl `powerauth-crypto` is built on Travis on all branches for now? I am just curious as this is the only one project with missing part
```
branches:
  only:
  - master
  - coverity_scan
```

Concerning the common configuration, there is a similar feature call but no progress so far https://github.com/travis-ci/travis-ci/issues/5385#issuecomment-169306763. Thus let's stick with the all-in-one-file approach.

romanstrobl:
The reason for `powerauth-crypto` being built on all branches is that it contains no PowerAuth dependencies, so it can always be built using Travis. The other projects may require publishing of artifacts. Plus it is really important that this project is buildable and tested.

PR #284, opened 2019-09-05
Fix #280: Added support for V3.1 PowerAuth signatures

### LINKED ISSUE #280: Crypto 3.1: Increase online signature entropy
In the next version of PowerAuth protocol, we're going to increase entropy of values used in online signatures. This change improves the information entropy in signatures and protection against brute force attacks.

### Proposed solution

The current format is:

- 8 digits, 10<sup>8</sup> combinations
- Example: 1FA: `12345678`, 2FA: `12345678-12345678`
- We'll keep this format for off-line signatures only.

The proposed new format will use Base64 encoding for :

- 16 bytes encoded in Base64, 10<sup>38</sup> combinations
- For 2FA and 3FA, use one Base64 encoded string for 32 or 48 bytes
- Example: 1FA: `MDEyMzQ1Njc4OWFiY2RlZg==`, 2FA: `MDEyMzQ1Njc4OWFiY2RlZjAxMjM0NTY3ODlhYmNkZWY=` 

How to construct such signature value:
```java
public String computeSignature(byte[] data, List<SecretKey> signatureKeys, byte[] CTR_DATA) {

    // ... compute signature components
    boolean isNumericFormat = isOfflineSignature || isOldFormat;
    String[] signatureComponents = new String[signatureKeys.size()];
    byte[] signatureBytes = new byte[signatureKeys.size() * 16];
    
    for (int i = 0; i < signatureKeys.size(); i++) {
        byte[] KEY_SIGNATURE = KeyConversion.secretKeyFromBytes(signatureKey.get(0));
        byte[] KEY_DERIVED = Mac.hmacSha256(KEY_SIGNATURE, CTR_DATA);

        // ... compute signature key using more than one keys, at most 2 extra keys
        // ... this skips the key with index 0 when i == 0
        for (int j = 0; j < i; j++) {
            KEY_SIGNATURE = KeyConversion.secretKeyFromBytes(signatureKey.get(j + 1));
            KEY_DERIVED_CURRENT = Mac.hmacSha256(KEY_SIGNATURE, CTR_DATA);
            KEY_DERIVED = Mac.hmacSha256(KEY_DERIVED_CURRENT, KEY_DERIVED);
        }

        // ... sign the data
        byte[] SIGNATURE_LONG = Mac.hmacSha256(KEY_DERIVED, DATA);

        if (isNumericFormat) {
            // offline signature, or old format ... decimalize the signature component
            int signComponent = (ByteUtils.getInt(ByteUtils.truncate(SIGNATURE_LONG, 4)) & 0x7FFFFFFF) % Math.pow(10,8);
            signatureComponents[i] = String.valueOf(signComponent);
        } else {
            // New online format ... Append last 16 bytes from SIGNATURE_LONG to array with signature bytes
            ByteUtils.copy(SIGNATURE_LONG, 16, signatureBytes, i * 16, 16);
        }
    }
    if (isNumericFormat) {
        // ... join the signature component using "-" character.
        return String.join("-", signatureComponents);   
    } else {
        // New online format
        return Base64.encode(signatureBytes);
    }
}
```

hvge:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/ECPublicKeyFingerprint.java
```diff
@@ -90,7 +90,7 @@ public static String compute(ECPublicKey devicePublicKey, ECPublicKey serverPubl
             }
             int index = hash.length - 4;
             int number = (ByteBuffer.wrap(hash).getInt(index) & 0x7FFFFFFF) % (int) (Math.pow(10, PowerAuthConfiguration.FINGERPRINT_LENGTH));
-            return String.format("%0" + PowerAuthConfiguration.SIGNATURE_LENGTH + "d", number);
+            return String.format("%0" + PowerAuthConfiguration.FINGERPRINT_LENGTH + "d", number);
```
hvge: This change fixes some previous copy-paste bug. Both constants have equal value, but with different meaning.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/signature/PowerAuthClientSignature.java
```diff
@@ -43,12 +44,13 @@
      * @param data Data to be signed.
      * @param signatureKeys A signature keys.
      * @param ctrData Hash based counter / index of the derived key KEY_DERIVED.
+     * @param signatureFormat Format of signature to produce.
```
romanstrobl: Nit pick - I think `compute` or `calculate` is better than `produce`.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -83,46 +85,118 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
-     * @return PowerAuth signature for given data.
+     * @return Decimal formatted PowerAuth signature for given data.
+     * @throws GenericCryptoException In case signature computation fails.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     */
+    private String computePowerAuthDecimalSignature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData) throws GenericCryptoException, CryptoProviderException {
+        // Prepare holder for signature components
+        final String[] signatureStringComponents = new String[signatureKeys.size()];
+        // Compute signature components
+        final byte[][] signatureComponents = computePowerAuthSignatureComponents(data, signatureKeys, ctrData);
```
romanstrobl: A style comment - for me `byte[][]` is a bit confusing and smells like C. I would personally prefer `List<byte[]>`.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -475,6 +476,7 @@ public void testSignatureValidationV2() throws Exception {
             KeyPair deviceKeyPair = keyGenerator.generateKeyPair();
             PrivateKey devicePrivateKey = deviceKeyPair.getPrivate();
 
+            final PowerAuthSignatureFormat signatureFormat = PowerAuthSignatureFormat.getFormatForSignatureVersion("2.0");
```
romanstrobl: I would prefer to use `2.1` as latest `2.x` version.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/signature/PowerAuthSignatureFormatTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void testValidVersions() throws Exception {
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("2.0"), PowerAuthSignatureFormat.DECIMAL);
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("2.1"), PowerAuthSignatureFormat.DECIMAL);
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("3.0"), PowerAuthSignatureFormat.DECIMAL);
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("3.1"), PowerAuthSignatureFormat.BASE64);
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("3.2"), PowerAuthSignatureFormat.BASE64);
+        assertEquals(PowerAuthSignatureFormat.getFormatForSignatureVersion("4.0"), PowerAuthSignatureFormat.BASE64);
+    }
+    @Test(expected = GenericCryptoException.class)
```
romanstrobl: Missing newline.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/signature/PowerAuthSignatureTest.java
```diff
@@ -315,7 +320,141 @@ public void testSignatureForDataV3() throws Exception {
...
+
+        KeyPair deviceKeyPair = keyGenerator.generateKeyPair();
+        PrivateKey devicePrivateKey = deviceKeyPair.getPrivate();
+        PublicKey devicePublicKey = deviceKeyPair.getPublic();
+
+        System.out.println("## Device Private Key: " + BaseEncoding.base64().encode(keyConvertor.convertPrivateKeyToBytes(devicePrivateKey)));
+        System.out.println("## Device Public Key:  " + BaseEncoding.base64().encode(keyConvertor.convertPublicKeyToBytes(devicePublicKey)));
+
+        final PowerAuthSignatureFormat signatureFormat = PowerAuthSignatureFormat.getFormatForSignatureVersion("3.1");
+        assertEquals(signatureFormat, PowerAuthSignatureFormat.BASE64);
```
romanstrobl: This is slightly incorrect. Expected value is first parameter in `assertEquals` method, so we should swap the parameters.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -83,46 +85,118 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
+     *
+     * @param data Data to be signed.
+     * @param signatureKeys Keys for computing the signature.
+     * @param ctrData Counter byte array / derived key index.
+     * @param format Format of signature to produce.
+     * @return PowerAuth signature for given data.
+     * @throws GenericCryptoException In case signature computation fails.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     */
+    public String computePowerAuthSignature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData, PowerAuthSignatureFormat format) throws GenericCryptoException, CryptoProviderException {
```
romanstrobl: Shouldn't we check input parameters in this method? I know this was not done before, but it is a good practice. We can check for nullability of all parameters and make sure `signatureKeys` list is not empty.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/signature/PowerAuthSignatureTest.java
```diff
@@ -94,6 +95,8 @@ public void testSignatureForDataV2() throws Exception {
         System.out.println("## Device Private Key: " + BaseEncoding.base64().encode(keyConvertor.convertPrivateKeyToBytes(devicePrivateKey)));
         System.out.println("## Device Public Key:  " + BaseEncoding.base64().encode(keyConvertor.convertPublicKeyToBytes(devicePublicKey)));
 
+        final PowerAuthSignatureFormat signatureFormat = PowerAuthSignatureFormat.getFormatForSignatureVersion("2.1");
+        assertEquals(signatureFormat, PowerAuthSignatureFormat.DECIMAL);
```
romanstrobl: Parameters are swapped, expected value is always first in unit tests.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/signature/PowerAuthSignatureTest.java
```diff
@@ -225,6 +228,8 @@ public void testSignatureForDataV3() throws Exception {
         System.out.println("## Device Private Key: " + BaseEncoding.base64().encode(keyConvertor.convertPrivateKeyToBytes(devicePrivateKey)));
         System.out.println("## Device Public Key:  " + BaseEncoding.base64().encode(keyConvertor.convertPublicKeyToBytes(devicePublicKey)));
 
+        final PowerAuthSignatureFormat signatureFormat = PowerAuthSignatureFormat.getFormatForSignatureVersion("3.0");
+        assertEquals(signatureFormat, PowerAuthSignatureFormat.DECIMAL);
```
romanstrobl: Swapped parameters.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
hvge: You're right. It was also for me a little bit confusing, but did not found a right replacement.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
hvge: This doesn't matter. I wanted to test that function for various inputs, but then I wrote a separate unit test which covers all variants.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -83,46 +85,118 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
+     */
+    private String computePowerAuthBase64Signature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData) throws GenericCryptoException, CryptoProviderException {
+        // Prepare array of bytes for a complete signature
+        final byte[] signatureBytes = new byte[signatureKeys.size() * PowerAuthConfiguration.SIGNATURE_BINARY_LENGTH];
+        // Compute signature components
+        final byte[][] signatureComponents = computePowerAuthSignatureComponents(data, signatureKeys, ctrData);
+        // Convert signature components into one Base64 encoded signature string
+        for (int i = 0; i < signatureComponents.length; i++) {
+            final byte[] signatureComponent = signatureComponents[i];
+            final int sourceOffset = signatureComponent.length - PowerAuthConfiguration.SIGNATURE_BINARY_LENGTH;
```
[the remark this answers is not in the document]
hvge: It's already tested in `computePowerAuthSignatureComponents()`. Both functions are private, so I don't see how can be this misused. I've added comment to that function.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/signature/PowerAuthSignatureFormatTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void testInvalidFormat1() throws Exception {
+        PowerAuthSignatureFormat.getFormatForSignatureVersion("3.1.1");
+    }
+
+    @Test(expected = GenericCryptoException.class)
+    public void testInvalidFormat2() throws Exception {
+        PowerAuthSignatureFormat.getFormatForSignatureVersion("foo");
+    }
+}
```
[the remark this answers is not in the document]
hvge: Right

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -648,7 +651,124 @@ public void testSignatureValidationV3() throws Exception {
...
+                    output.put("signature", signature);
+                    testSet.addData(input, output);
+
+                    ctrData = hashBasedCounter.next(ctrData);
+                }
+
+                for (int k = 0; k < signatureCount; k++) {
+
+                    // generate random data
+                    byte[] data = keyGenerator.generateRandomBytes((int) (Math.random() * dataMax));
```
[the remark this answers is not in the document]
hvge: Do we have a real problem with this? I don't want to overcomplicate things when there's no real issue.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
hvge: Right, but I think that it's OK to have `null` data. That's acceptable in HMAC calculation and means an empty data.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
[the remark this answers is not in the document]
romanstrobl: No real issue here, you can ignore it. It just looks a bit weird because I am used to using `SecureRandom` all the time, I would expect no usage of Math.random() in crypto library because it is weak.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
romanstrobl: Yes, but `null` `ctrData`, `format` or `signatureKeys` is not so cool.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -83,46 +85,118 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
-        HMACHashUtilities hmac = new HMACHashUtilities();
+        final HMACHashUtilities hmac = new HMACHashUtilities();
 
-        // Prepare holder for signature components
-        String[] signatureComponents = new String[signatureKeys.size()];
+        // Prepare array for signature binary components.
+        final byte[][] signatureComponents = new byte[signatureKeys.size()][];
 
-        CryptoProviderUtil keyConvertor = PowerAuthConfiguration.INSTANCE.getKeyConvertor();
+        final CryptoProviderUtil keyConverter = PowerAuthConfiguration.INSTANCE.getKeyConvertor();
```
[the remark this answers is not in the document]
hvge: In that case, we should throw `GenericCryptoException` directly from `getKeyConvertor()` function. It makes no sense to throw NPE, catch that in all possible situations and re-throw as another exception.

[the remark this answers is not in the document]
romanstrobl: Ok, I will look into this later as part of another task, so you can ignore it for now.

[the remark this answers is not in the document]
hvge: OK, I'll ignore this request. The purpose of this change is to bring a support for 3.1 signatures and not to refactor the whole stack :) I've tried, but failed...

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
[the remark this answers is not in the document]
hvge: OK

PR #288, opened 2019-09-12
Fix #280: Signature format validator now supports Base64 signatures.

### LINKED ISSUE #280: Crypto 3.1: Increase online signature entropy

hvge:
This change also contains refactoring for `PowerAuthSignatureFormat.getFormatForSignatureVersion()` method and adds new functionality to that enumeration.

### DIFF
--- a/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureFormat.java
+++ b/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureFormat.java
@@ -40,7 +40,7 @@ public enum PowerAuthSignatureFormat {
      *     <li>For all versions of offline signatures.</li>
      * </ul>
      */
-    DECIMAL("DECIMAL"),
+    DECIMAL,
     /**
      * Each signature's factor is represented by 16-bytes long binary data. If the signature is composed from more than
      * one factor, then the binary sequences are concatenated one after another. The whole signature is then represented
@@ -51,16 +51,7 @@ public enum PowerAuthSignatureFormat {
      * </ul>
      * This type of formatting is currently used for {@code 3.1} version of online signatures and newer.
      */
-    BASE64("BASE64");
-
-    /**
-     * String representation associated with the enumeration.
-     */
-    private String value;
-
-    PowerAuthSignatureFormat(final String value) {
-        this.value = value;
-    }
+    BASE64;
 
     /**
      * Map that translates string into {@link PowerAuthSignatureFormat} enumeration.
@@ -75,7 +66,7 @@ public enum PowerAuthSignatureFormat {
     static {
         // Prepare string to enumeration mapping
         for (PowerAuthSignatureFormat format : PowerAuthSignatureFormat.values()) {
-            stringToEnumMap.put(format.value, format);
+            stringToEnumMap.put(format.toString(), format);
         }
         // Prepare version to enumeration mapping
         versionToEnumMap.put("2.0", DECIMAL);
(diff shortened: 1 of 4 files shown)


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureFormat.java
```diff
@@ -37,7 +40,7 @@
      *     <li>For all versions of offline signatures.</li>
      * </ul>
      */
-    DECIMAL,
+    DECIMAL("DECIMAL"),
```
romanstrobl: Isn't the value a bit duplicate when you can get the same value using `toString()`?

hvge: OK

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureTypes.java
```diff
@@ -77,12 +77,14 @@
...
-        } else {
-            return type;
+        if (value != null) {
+            final PowerAuthSignatureTypes type = map.get(value.toLowerCase());
+            if (type != null) {
+                return type;
+            }
         }
+        // Otherwise try to guess the most usual suspect...
+        return PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE;
```
romanstrobl: What if some uses a signature type with a typo, it will fall back to `POSSESSION_KNOWLEDGE` and then work magically? Shouldn't we rather prevent this case?

Issue #291, opened 2019-09-23
Crypto 3.1: Document protocol changes


hvge:
We have to update our "crypto" documentation with all changes introduced in V3.1 protocol. 

Related to #276

PR #292, opened 2019-09-28
Features/crypto 31

hvge:
This change adds support for PowerAuth protocol 3.1 to develop branch.


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureTypes.java
```diff
@@ -77,12 +77,14 @@
...
-        } else {
-            return type;
+        if (value != null) {
+            final PowerAuthSignatureTypes type = map.get(value.toLowerCase());
+            if (type != null) {
+                return type;
+            }
         }
+        // Otherwise try to guess the most usual suspect...
+        return PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE;
```
romanstrobl: I am not sure if this is a good idea. This way we will allow mistyped signature types which will magically work as `POSSESSION_KNOWLEDGE`. It is better to throw an exception IMO.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyDerivationUtils.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param keyGenerator {@link KeyGenerator} instance to be used in the class.
+     */
+    public KeyDerivationUtils(KeyGenerator keyGenerator) {
+        this.keyGenerator = keyGenerator;
+    }
+
+    /**
+     * Derivation index used to derive KEY_TRANSPORT_IV from KEY_TRANSPORT
+     */
+    private static final long STATUS_BLOB_TRANSPORT_IV_INDEX = 3000;
```
romanstrobl: Shouldn't we document this in our developers portal? The value seems a bit magic here.

powerauth-java-http/src/main/java/io/getlime/security/powerauth/http/validator/ValueTypeValidator.java
```diff
@@ -97,7 +97,26 @@ public static boolean isValidSignatureType(String signatureType) {
...
+                case 17:
+                case 26:
+                    // "2.0", "2.1", "3.0" signature version uses "DECIMAL" format
+                    return signature.matches(signatureRegex);
+                case 24:
+                case 44:
+                case 64:
+                    // "3.1" and later signatures uses "BASE64" format.
+                    // We don't need to validate an exact number of encoded bytes. This is due to fact,
+                    // that if input string length can only be 24, 44 or 64, then the encoded output length
```
romanstrobl: Shouldn't the `encoded` word be `decoded`? We are checking length of string after the `base64.decode()` call.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -83,46 +86,131 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
-    public String computePowerAuthSignature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData) throws GenericCryptoException, CryptoProviderException {
+    private List<byte[]> computePowerAuthSignatureComponents(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData) throws GenericCryptoException, CryptoProviderException {
         // Prepare a hash
-        HMACHashUtilities hmac = new HMACHashUtilities();
+        final HMACHashUtilities hmac = new HMACHashUtilities();
 
-        // Prepare holder for signature components
-        String[] signatureComponents = new String[signatureKeys.size()];
+        // Prepare array for signature binary components.
+        final ArrayList<byte[]> signatureComponents = new ArrayList<byte[]>();
```
romanstrobl: `final ArrayList<byte[]> signatureComponents = new ArrayList<byte[]>();` hurts Java developer's eyes. Let's use `final List<byte[]> signatureComponents = new ArrayList<>();`

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyDerivationUtils.java
hvge: See #291

powerauth-java-http/src/main/java/io/getlime/security/powerauth/http/validator/ValueTypeValidator.java
hvge: It's only an assumption that several string lengths are in direct relationship to an expected number of encoded bytes (someone encoded that bytes before). We're going to validate that expected number of encoded bytes. If it's faster, then I would use regexpt to validate that string, instead of full decode. From this point of view, it doesn't matter that the function actually decodes that bytes to validate the lengths.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
hvge: Right

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/enums/PowerAuthSignatureTypes.java
hvge: I agree with you, but it's an old code. I would change this in a separate refactoring, because adding an exception here will break many other things (in restful integration for example).

romanstrobl: Ok, let's raise a ticket.

PR #295, opened 2019-10-18
Issues/291 crypto31 doc

hvge:
This PR adds crypto 3.1 documentation and missing counters (in signature validation) behavior to docs.


docs/Activation-Status.md
```diff
@@ -4,18 +4,43 @@ PowerAuth Client may need to check for an activation status, so that it can dete
...
 
 Following sequence diagram shows the activation status check in more detail.
 
 ![Check Activation Status](./resources/images/sequence_activation_status.png)
 
+## Status Blob Encryption
+
+1. Both, client and server calculate `KEY_TRANSPORT_IV` as:
+    ```java
+    SecretKey KEY_TRANSPORT_IV = KDF.derive(KEY_TRANSPORT, 3000)
```
romanstrobl: Perhaps we should add an explanation why value `3000` is used?

docs/Computing-and-Validating-Signatures.md
```diff
@@ -97,6 +154,31 @@ _Note: Note that the `APPLICATION_SECRET` is technically outside the request dat
...
+  - It's randomly initialized and exchanged during the activation, or in the protocol upgrade process.
+  - In protocol version `3.1` client can synchronize its counter with the server. 
+- `CTR_LOOK_AHEAD`, tolerance set on server to overcome ahead clients
+  - Server is trying to calculate and validate the signature ahead in time, in half-closed interval defined by this tolerance: `[CTR, CTR + CTR_LOOK_AHEAD)`.
+  - Default value is `20`
+- `FAILED_ATTEMPTS`, how many attempts failed before in row
+   - Initial value is 0.
+   - If value reaches value defined in `MAX_FAILED_ATTEMPTS`, then activation is set to `BLOCKED` state.
+   - Value is increased in case that signature validation fails (see description below)
+- `MAX_FAILED_ATTEMPTS`, how is maximum failed attempts in row that ends in blocked activation.
```
romanstrobl: I do not understand this sentence (the `in row` part). After reading it 3 times, I think I start to understand... Perhaps it is clearer as `how many maximum failed attempts in a row result in blocked activation`.

docs/Computing-and-Validating-Signatures.md
```diff
@@ -126,4 +208,18 @@ for (CRT_ITER = CTR; CTR_ITER++; CRT_ITER < CRT + TOLERANCE) {
 return VERIFIED;
 ```
 
-Additionally, server may implement partial signature validation - basically evaluate each signature component separately. This may be used to determine if a failed attempt counter should be incremented or not (since this allows distinguishing attacker who has a physical access to the PowerAuth Client from attacker who randomly guesses signature).
+#### Success
+
+In case that signature is verified, then:
```
romanstrobl: I would prefer to write `In case that signature is successfully verified`.

docs/Activation-Status.md
hvge: The constant is explained in the list of used keys. Just like other similar constants.

romanstrobl: Ok, I see now.

PR #313, opened 2019-12-03
Fix #308: Plain CTR_DATA is no longer in status.

### LINKED ISSUE #308: Crypto 3.1: Do not send CTR_DATA in status
The status blob should contain only a hash calculated from `CTR_DATA`, instead of the direct value of the counter:
```java
byte[] CalculateCounterHash(byte[] CTR_DATA) {
    SecretKey KEY_TRANSPORT_CTR = KDF.derive(KEY_TRANSPORT, 4000);
    byte[] CTR_DATA_HASH = KeyConversion.getBytes(KDF_INTERNAL.derive(KEY_TRANSPORT_CTR, CTR_DATA));
    return CTR_DATA_HASH;
}
```

Unfortunately, this change makes the current counter synchronization on mobile clients ineffective. To fix this, we should enhance the status blob with a limited information about the actual counter. Let's call it `CTR_INFO`:

```java
byte CTR_INFO = (byte) (CTR & 0xFF)
```

The mobile client then can synchronize its counter with using the following algorithm:

1. Initial values
   ```java
   byte   local_CTR_INFO = persistentData.CTR_INFO;
   byte[] local_CTR_DATA = persistentData.CTR_DATA;
   byte[] local_CTR_DATA_HASH = CalculateCounterHash(local_CTR_DATA);
   byte   server_CTR_INFO      = statusBlob.CTR_INFO;
   byte[] server_CTR_DATA_HASH = statusBlob.CTR_DATA_HASH;
   ```

1. Determine whether counter is already synchronized
   ```java
   if (local_CTR_INFO == server_CTR_INFO && 
       local_CTR_DATA_HASH.equals(server_CTR_DATA_HASH)) {
       // counter is synchronized, do nothing else...
       return OK;
   }
   ```

1. Determine whether server is ahead:
   ```java
   // This is equal to current impl. We're trying to move local counter close to the server's, 
   // but instead of comparing CTR_DATA, we have to compare CTR_DATA_HASH
   int attempts = 0;
   while (true) {
       // calculate next CTR_DATA and CTR_INFO
       local_CTR_DATA = SHA256(local_CTR_DATA);
       local_CTR_INFO = local_CTR_INFO + 1;
       // calculate next local_CTR_DATA_HASH
       local_CTR_DATA_HASH = CalculateCounterHash(local_CTR_DATA);
       if (local_CTR_DATA_HASH.equals(server_CTR_DATA_HASH) && 
           local_CTR_INFO == server_CTR_INFO) {
           // Hash & Counter equals
           // Keep local_CTR_INFO and local_CTR_DATA in persistent data
           return OK;
       }
       if (++attempts == statusBlob.CTR_LOOK_AHEAD) {
           // We cannot match 
           break;
       }
   }
   ```
   
1. Determine whether client is ahead:
   ```cpp
   // Note that byte is unsigned char. 
   int CalculateCounterDistance(byte local, byte server)
   {
       int L = local;
       int S = server;
       // Calculate possible distances
       int d1 = L - S;
       int d2 = 256 + L - S;
       int d3 = L - (256 + S);
       // Find minimum absolute distance from possible distances
       int d1a = abs(d1);
       int d2a = abs(d2);
       int d3a = abs(d3);
       int distance_abs = min(d1a, min(d2a, d3a));
       // Determine which one is it.
       if (distance_abs == d1a) {
           return d1;
       } else if (distance_abs == d2a) {
           return d2;
       }
       return d3;
   }
   
   int distance = CalculateCounterDistance(local_CTR_INFO, server_CTR_INFO);
   if (distance > 0 && distance < statusBlob.CTR_LOOK_AHEAD) {
       // Client can be synchronized by calculating signature on the server.
       return CALCULATE_SIGN;
   }
   // Client cannot be synchronized.
   return DEADLOCK;
   ```

hvge:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/HashBasedCounterUtils.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param expectedCtrData Expected hash based counter.
+     * @param transportKey Transport key.
+     * @return {@code true} in case that received hash equals to hash calculated from counter data.
+     * @throws InvalidKeyException When invalid key is provided.
+     * @throws GenericCryptoException In case key derivation fails.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     */
+    public boolean verifyHashForHasBasedCounter(byte[] receivedCtrDataHash, byte[] expectedCtrData, SecretKey transportKey)
+            throws CryptoProviderException, InvalidKeyException, GenericCryptoException {
+        if (expectedCtrData == null || expectedCtrData.length != CTR_DATA_LENGTH) {
```
romanstrobl: Let's verify that `transportKey` is not null.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/HashBasedCounterUtils.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param ctrData Hash-based counter.
+     * @param transportKey Transport key.
+     * @return Hash calculated from provided hash-based counter.
+     * @throws GenericCryptoException In case that key derivation fails or you provided invalid ctrData.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     * @throws InvalidKeyException In case that transport key is not valid.
+     */
+    public byte[] calculateHashFromHashBasedCounter(byte[] ctrData, SecretKey transportKey)
+            throws CryptoProviderException, InvalidKeyException, GenericCryptoException {
+        if (ctrData == null || ctrData.length != CTR_DATA_LENGTH) {
```
romanstrobl: Let's also verify that `transportKey` is not null.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/HashBasedCounterUtils.java
```diff
@@ -0,0 +1,110 @@
...
+ * and verification of hash based counter over the network.
+ */
+public class HashBasedCounterUtils {
+
+    private final KeyGenerator keyGenerator;
+
+    /**
+     * @param keyGenerator {@link KeyGenerator} instance to be used in the class.
+     */
+    public HashBasedCounterUtils(KeyGenerator keyGenerator) {
```
romanstrobl: Do we really need to pass `KeyGenerator` as parameter? We can just use `keyGenerator = new KeyGenerator` in field initialization, there is no state required to be passed and creating new instance is cheap.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/model/ActivationStatusBlobInfo.java
```diff
@@ -150,18 +151,34 @@ public void setCtrLookAhead(byte ctrLookAhead) {
...
 
     /**
-     * Get counter data.
-     * @return Counter data.
+     * Get least significant byte from the counter.
+     * @return Least significant byte from the counter.
      */
-    public byte[] getCtrData() {
-        return ctrData;
+    public byte getCtrInfo() {
```
romanstrobl: I do not like the `info` word in the method name, it is quite generic. Perhaps we can use something like `getCtrMasked` or `getCtrByte`... maybe you have a better idea.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/HashBasedCounterUtils.java
hvge: Removed also from previously added `KeyDerivationUtils`

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/model/ActivationStatusBlobInfo.java
hvge: Renamed to `ctrByte` (e.g. `getCtrByte`).

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
```diff
@@ -300,9 +303,26 @@ public ActivationStatusBlobInfo getStatusFromEncryptedBlob(byte[] cStatusBlob, b
...
+     *
+     * @param receivedCtrDataHash Value received from the server, containing hash, calculated from hash based counter.
+     * @param expectedCtrData Expected hash based counter.
+     * @param transportKey Transport key.
+     * @return {@code true} in case that received hash equals to hash calculated from counter data.
+     * @throws InvalidKeyException When invalid key is provided.
+     * @throws GenericCryptoException In case key derivation fails.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     */
+    public boolean verifyHashForHasBasedCounter(byte[] receivedCtrDataHash, byte[] expectedCtrData, SecretKey transportKey)
```
romanstrobl: Typo: `verifyHashForHasBasedCounter`

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
```diff
@@ -300,9 +303,26 @@ public ActivationStatusBlobInfo getStatusFromEncryptedBlob(byte[] cStatusBlob, b
...
+            throws CryptoProviderException, InvalidKeyException, GenericCryptoException {
+        return new HashBasedCounterUtils().verifyHashForHasBasedCounter(receivedCtrDataHash, expectedCtrData, transportKey);
```
romanstrobl: Typo: `verifyHashForHasBasedCounter`

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/HashBasedCounterUtils.java
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * @param receivedCtrDataHash Value received from the server, containing hash, calculated from hash based counter.
+     * @param expectedCtrData Expected hash based counter.
+     * @param transportKey Transport key.
+     * @return {@code true} in case that received hash equals to hash calculated from counter data.
+     * @throws InvalidKeyException When invalid key is provided.
+     * @throws GenericCryptoException In case key derivation fails.
+     * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
+     */
+    public boolean verifyHashForHasBasedCounter(byte[] receivedCtrDataHash, byte[] expectedCtrData, SecretKey transportKey)
```
romanstrobl: Typo: `verifyHashForHasBasedCounter`

PR #320, opened 2019-12-12
Fix #147: Offline signatures docs

### LINKED ISSUE #147: Document offline signatures
We are missing documentation on offline signature principle.

hvge:


docs/Computing-and-Validating-Signatures.md
```diff
@@ -152,6 +152,11 @@ _Note: Note that the `APPLICATION_SECRET` is technically outside the request dat
 	- In case of request with body (such as POST and PUT requests), data from the resource body (bytes) are encoded using Base64 with UTF-8 encoding and appended:
 		- `REQUEST_DATA = BASE64.encode(ByteUtils.getBytes(HTTP['body']))`
 
+In case of data for offline signature is being normalized, then the following exceptions are in place:
```
romanstrobl: I suggest to change the sentence a bit (it is a bit unclear what an exception is):

```
In case the data for offline signature is being normalized, then the following rules are applied:
```

PR #340, opened 2020-05-18
Fix #339: Switch RandomGenerator to Bouncy Castle

### LINKED ISSUE #339: Switch RandomGenerator to Bouncy Castle
The Bouncy Castle `SecureRandom` implementation provides better security:
- a NIST-approved `Hash_DRBG` algorithm based on `SHA-512` is used
- the random generator reseeds after each request
- the random generator is not dependent on OS and Java version
- the algorithm does not have known quantum attacks

The only downside is lower performance, however it can be resolved by adding CPU cores if it even becomes an issue.

Let's switch to the Bouncy Castle provider for SecureRandom.

romanstrobl:
This pull requests switches the random generator to Bouncy Castle `DEFAULT` algorithm, which is currently `Hash_DRBG` with `SHA-512`.

Most of the pull request is about exception handling, because the `SecureRandom` needs to be created after Bouncy Castle is initialized which is typically done by `Security.addProvider(new BouncyCastleProvider());`.

### DIFF
--- a/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
+++ b/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
@@ -45,8 +45,16 @@ import java.util.Arrays;
 public class PowerAuthClientActivation {
 
     private final SignatureUtils signatureUtils = new SignatureUtils();
-    private final KeyGenerator keyGenerator = new KeyGenerator();
     private final KeyConvertor keyConvertor = new KeyConvertor();
+    private final KeyGenerator keyGenerator;
+
+    /**
+     * PowerAuth client activation constructor.
+     * @throws CryptoProviderException In case key cryptography provider is incorrectly initialized.
+     */
+    public PowerAuthClientActivation() throws CryptoProviderException {
+        keyGenerator = new KeyGenerator();
+    }
 
     /**
      * Verify the signature of activation code using Master Public Key.
--- a/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
+++ b/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
@@ -45,16 +45,8 @@ import java.util.Arrays;
 public class PowerAuthClientActivation {
 
     private final SignatureUtils signatureUtils = new SignatureUtils();
+    private final KeyGenerator keyGenerator = new KeyGenerator();
     private final KeyConvertor keyConvertor = new KeyConvertor();
-    private final KeyGenerator keyGenerator;
-
-    /**
-     * PowerAuth client activation constructor.
-     * @throws CryptoProviderException In case key cryptography provider is incorrectly initialized.
-     */
-    public PowerAuthClientActivation() throws CryptoProviderException {
-        keyGenerator = new KeyGenerator();
-    }
 
     /**
      * Verify the signature of activation code using Master Public Key.
@@ -86,8 +78,9 @@ public class PowerAuthClientActivation {
      * Generate a new activation nonce.
      *
      * @return A new activation nonce.
+     * @throws CryptoProviderException In case key cryptography provider is incorrectly initialized.
      */
-    public byte[] generateActivationNonce() {
+    public byte[] generateActivationNonce() throws CryptoProviderException {
         return keyGenerator.generateRandomBytes(16);
     }
 
--- a/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/NonPersonalizedEncryptor.java
+++ b/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/NonPersonalizedEncryptor.java
@@ -53,9 +53,9 @@ public class NonPersonalizedEncryptor {
 
     // Create new working objects
     private final AESEncryptionUtils aes = new AESEncryptionUtils();
+    private final KeyGenerator generator = new KeyGenerator();
     private final HMACHashUtilities hmac = new HMACHashUtilities();
     private final KeyConvertor keyConvertor = new KeyConvertor();
-    private KeyGenerator generator;
 
     /**
      * Create a new encryptor using provided applicationKey, application master server public key and session index.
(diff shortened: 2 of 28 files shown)


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
```diff
@@ -45,8 +45,16 @@
...
-    private final KeyGenerator keyGenerator = new KeyGenerator();
     private final KeyConvertor keyConvertor = new KeyConvertor();
+    private final KeyGenerator keyGenerator;
+
+    /**
+     * PowerAuth client activation constructor.
+     * @throws CryptoProviderException In case key cryptography provider is incorrectly initialized.
+     */
+    public PowerAuthClientActivation() throws CryptoProviderException {
+        keyGenerator = new KeyGenerator();
```
petrdvorak, line 56: It seems a bit weird we only initialize `keyGenerator` in the constructor - is there a way to use the inline field initialization as we do for `keyConveror`? Maybe making a `CryptoProviderException` runtime?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/NonPersonalizedEncryptor.java
```diff
@@ -53,22 +53,24 @@
...
      * @param ephemeralPublicKeyString Ephemeral public key
+     * @throws CryptoProviderException In case key cryptography provider is incorrectly initialized.
      */
-    public NonPersonalizedEncryptor(byte[] applicationKey, byte[] sessionRelatedSecretKey, byte[] sessionIndex, byte[] ephemeralPublicKeyString) {
+    public NonPersonalizedEncryptor(byte[] applicationKey, byte[] sessionRelatedSecretKey, byte[] sessionIndex, byte[] ephemeralPublicKeyString) throws CryptoProviderException {
         this.applicationKey = applicationKey;
         this.sessionIndex = sessionIndex;
         this.sessionRelatedSecretKey = sessionRelatedSecretKey;
         this.ephemeralPublicKey = ephemeralPublicKeyString;
+        this.generator = new KeyGenerator();
```
petrdvorak, line 73: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesEncryptor.java
```diff
@@ -190,6 +190,7 @@ private EciesCryptogram encrypt(byte[] data, boolean useIv) throws EciesExceptio
             final byte[] iv;
             if (useIv) {
                 // V3.1+, generate random nonce and calculate IV
+                keyGenerator = new KeyGenerator();
```
petrdvorak, line 193: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesEnvelopeKey.java
```diff
@@ -85,6 +86,9 @@ public EciesEnvelopeKey(byte[] secretKey, byte[] ephemeralPublicKey) {
      */
     static EciesEnvelopeKey fromPublicKey(PublicKey publicKey, byte[] sharedInfo1) throws EciesException {
         try {
+            if (keyGenerator == null) {
```
petrdvorak, line 89: ... this looks a bit weird.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesEnvelopeKey.java
```diff
@@ -121,6 +125,9 @@ static EciesEnvelopeKey fromPublicKey(PublicKey publicKey, byte[] sharedInfo1) t
      */
     static EciesEnvelopeKey fromPrivateKey(PrivateKey ephemeralKeyPrivate, byte[] ephemeralPublicKeyBytes, byte[] sharedInfo1) throws EciesException {
         try {
+            if (keyGenerator == null) {
```
petrdvorak, line 128: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesEnvelopeKey.java
```diff
@@ -201,6 +208,9 @@ static EciesEnvelopeKey fromPrivateKey(PrivateKey ephemeralKeyPrivate, byte[] ep
             throw new EciesException("Nonce for IV derivation is not valid");
         }
         try {
+            if (keyGenerator == null) {
```
petrdvorak, line 211: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/client/activation/PowerAuthClientActivation.java
romanstrobl: I do not think this is a good idea. The checked exception tells the client code that this issue can happen and we know it can happen quite easily when BC is not set up correctly. It is better to let the client handle it including proper logging (and other cleanup if needed) than interrupting the processing unexpectedly.

romanstrobl:
Ok... if we want to keep the current style of instance construction, I have another proposal:
- we can handle the Bouncy Castle `SecureRandom.getInstance()` Exception in constructor and keep `random` instance `null`
- we can do the error handling in `generateRandomBytes` method instead and handle the case when `SecureRandom` was not set up correctly

romanstrobl:
The reason why field initialization is problematic is that we need to make sure `Security.addProvider(new BouncyCastleProvider())` is called before any field initialization involving the RandomGenerator is triggered, otherwise the correct `SecureRandom` is missing. I think we don't want to just fallback to Java SecureRandom, because this behaviour could depend on how Bouncy Castle library is initialized. I guess we can initialize the SecureRandom in a lazy way and this way we can keep the field initialization as is. I will prepare a proposal in a separate branch and we can choose the lesser evil.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesEnvelopeKey.java
romanstrobl: Yes, I know, it is caused by static context. Anyway I think we can handle all these style issues by initializing the SecureRandom lazily.

romanstrobl:
I think the current code looks much cleaner, so I pushed it to this branch. I made the `SecureRandom` initialization lazy, so that inline fields can be initialized without problems and the Bouncy Castle provider is safely registered before any `SecureRandom` initialization. We can compare the two possible solutions. 

But if you have an idea how to handle the initialization without any "lesser evil" option (meaning that `SecureRandom` initialization would stay in the `KeyGenerator` constructor), let me know.

PR #352, opened 2020-11-09
Fix #351: Missing volatile keyword in double checked locking

### LINKED ISSUE #351: Missing volatile keyword in double checked locking
The `volatile` keyword should be added into `KeyGenerator` class for the `random` field to avoid lock evasion issues when the class is used from multiple threads.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #364, opened 2021-03-17
Fix #363: Increase Argon2 salt size

### LINKED ISSUE #363: Increase Argon2 salt size
Quote: "16 bytes is sufficient for all applications, but can be reduced to 8 bytes in the case of space constraints".

There are no space constraints, so let's use 16 bytes.

romanstrobl:


hvge (approved):
Looks OK

PR #368, opened 2021-05-03
Fix #367: Perform full point validation during import

### LINKED ISSUE #367: Perform full point validation during import
See: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf, chapter 5.6.2.3.3.

romanstrobl:
I added two public key validations:
- point at infinity not allowed as public key
- calculation of point order and its verification against order defined in EC curve


petrdvorak (approved):
Looks OK

PR #369, opened 2021-05-05
Add a test for invalid curve order simulating an EC curve fault attack

romanstrobl:
I added a test for point order validation simulating an EC curve fault attack by altering the curve order.


petrdvorak (approved):
Looks OK

PR #385, opened 2021-11-24
Fix #384: Use BC library for array comparison

### LINKED ISSUE #384: Use BC library for array comparison
The BC library provides an implementation of array utilities in the `Arrays` class.

romanstrobl:
Migration to BC library due to constant time array comparisons.


petrdvorak (approved):
Looks OK

PR #399, opened 2022-04-13
Fix #398: Improve migration guide for release 1.2.5

### LINKED ISSUE #398: Improve migration guide for release 1.2.5
We should improve the migration guide because of release `1.2.5` which requires extra DB migration for PowerAuth server.

romanstrobl:
Separation of releases `2021.11` and `2022.04` for better explanation of DB migration in mid-release.

See also:
https://github.com/wultra/powerauth-server/pull/683


petrdvorak (approved):
Looks good

PR #401, opened 2022-06-07
Fix #400: Update dependencies

romanstrobl:

PR #402, opened 2022-06-08
Improve maven dependency management

banterCZ:
A suggestion how to improve maven dependency management as a follow-up to https://github.com/wultra/powerauth-crypto/pull/401

### DIFF
--- a/powerauth-java-crypto/pom.xml
+++ b/powerauth-java-crypto/pom.xml
@@ -28,24 +28,20 @@
 		<groupId>io.getlime.security</groupId>
 		<artifactId>powerauth-crypto-parent</artifactId>
 		<version>1.3.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


powerauth-java-crypto/pom.xml
```diff
@@ -28,24 +28,20 @@
 		<groupId>io.getlime.security</groupId>
 		<artifactId>powerauth-crypto-parent</artifactId>
 		<version>1.3.0-SNAPSHOT</version>
-		<relativePath>../pom.xml</relativePath>
```
banterCZ, line 31: This is default value.

petrdvorak (approved):
Again, I see quite a low benefit of this change... but why not.

romanstrobl (approved):
Nice cleanup.

PR #414, opened 2022-09-16
Fix #413: Prepare Github Action for build and publishing artifact

### LINKED ISSUE #413: Prepare Github Action for build and publishing artifact

petrdvorak:


petrdvorak:
Using the Github actions for building Java artifacts seems pretty straight forward. Let's discuss this on next grooming. cc @zcgandcomp

petrdvorak:
Also, we can set everything up so that we automatically publish snapshot version to either Maven central, or Githubs own package repository. I think the 2nd option might have some advantages in our case, but I have seen it is simple to publish artifacts to both places.

banterCZ (approved):
Technically, it looks good. But as you mentioned, do we really need to have both repositories?

romanstrobl (approved):
Ok, but I am a bit unsure how we plan to store the GPG signing key and credentials for the publishing to Maven repo, see e.g.: https://github.com/jonashackt/github-actions-release-maven#create-encrypted-variables-in-github-actions

PR #415, opened 2022-10-05
Configure deployment to OSS Sonatype repository

romanstrobl:


petrdvorak (approved):

.github/workflows/maven.yml
```diff
@@ -35,4 +38,8 @@ jobs:
       with:
         name: PowerAuth Java HTTP Utilities
         path: powerauth-java-http
-
+    - name: Publish Libraries to OSS Sonatype Repository
+      run: mvn -B deploy --file pom.xml
```
banterCZ, line 42: What about to omit default values, such as `--file pom.xml`, but add `-U`?

banterCZ: Moreover, when we run `deploy`, `package` is executed again, isn't it?

PR #417, opened 2022-10-05
Fix #416: Visualize JUnit Tests Better

### LINKED ISSUE #416: Visualize JUnit Tests Better

petrdvorak:


.github/workflows/maven-test.yml
```diff
@@ -18,5 +18,15 @@ jobs:
         java-version: '11'
         distribution: 'temurin'
         cache: maven
-    - name: Run the Maven verify phase
-      run: mvn -B --update-snapshots verify
+    - name: Run Maven Build Step
+      run: mvn -B --update-snapshots -DskipTests package
+    - name: Run Maven Test Step
+      run: mvn -B --update-snapshots -Dmaven.test.failure.ignore=true test
```
banterCZ, line 24: Most probably I do not get GitHub actions yet. Isn't it too complicated do run `package` and then `test`? I guest that it runs `compile` twice.

petrdvorak: We need the package phase for this to work, we do not want to have it in one command with `test` since this would mix the results in one steps and be hard to combine with the required flags.

.github/workflows/maven-test.yml
```diff
@@ -18,5 +18,15 @@ jobs:
         java-version: '11'
         distribution: 'temurin'
         cache: maven
-    - name: Run the Maven verify phase
-      run: mvn -B --update-snapshots verify
+    - name: Run Maven Build Step
+      run: mvn -B -U -DskipTests package
+    - name: Run Maven Test Step
```
romanstrobl, line 23: Why do we run maven twice, once without tests and once with tests?

romanstrobl (approved):
Ok, but I am not sure why we run Maven twice if it can probably be just one step.

.github/workflows/maven-test.yml
petrdvorak: Does not work in one step - for publishing the report, this is the only way it works.

PR #421, opened 2022-10-06
Update maven-test.yml

petrdvorak:


romanstrobl (approved):
Ok, let's squash-merge these commits to avoid too many commits in history.

PR #425, opened 2022-10-06
Fix #424: GitHub action creates staging directory but does not use it

### LINKED ISSUE #424: GitHub action creates staging directory but does not use it
The GitHub action creates staging directory but does not use it for deployment:
https://github.com/wultra/powerauth-crypto/blob/develop/.github/workflows/maven-publish.yml#L30

romanstrobl:


petrdvorak (approved):
Looks OK

PR #432, opened 2022-10-10
Issues/431 reusable actions

zcgandcomp:
Please look at the structure and how it calls wultra-infrastructure.
I will erase testing secret job before moving from WIP to finished PR.

The intention is to run (compile and) test on merged code in protected branches and on any pull request leading to protected branches

The maven publishing is manually executed, but allowed only on protected branches


.github/workflows/maven-test.yml
```diff
@@ -2,28 +2,17 @@ name: Test with Maven
...
+    branches:
+     - 'develop'
+     - 'master'
+     - 'releases/**'
   pull_request:
-    branches: [ "develop" ]
+    branches:
+      - 'develop'
+      - 'master'
+      - 'releases/**'
```
romanstrobl, line 14: Note that the mobile devs use `release` name for release branches instead of `releases`:
https://github.com/wultra/powerauth-mobile-sdk/branches

In new repositories, `master` branch will be replaced by `main` branch.

Not a problem at all because we can configure branch names easily, just pointing out the little differences.

romanstrobl (approved):
I looked at the workflow and it seems fine for the basic usage.

We will need to add one more server ID for staging of release libraries to Maven central, see:

```
    <distributionManagement>
        <snapshotRepository>
            <id>ossrh-snapshots-distribution</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
        </snapshotRepository>
        <repository>
            <id>ossrh-staging-distribution</id>
            <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
        </repository>
    </distributionManagement>
```

This can be done later.

banterCZ (approved):

PR #433, opened 2022-10-13
Test/ci

zcgandcomp:
new structure for wultra/tasklist#23. change of pom.xml to allow profiling oss repo vs internal jfrog repo. 
change to new reusable prototype
preparation for manual triggering


.github/workflows/maven-deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+
+on:
+  workflow_dispatch:
+    branches:
+      - 'develop'
+      - 'master'
+      - 'release/*'
+      - 'test/ci'
+    inputs:
+      style:
```
romanstrobl, line 11: I am not sure `style` is the right word for this. I would prefer `target` or `release_type` or something similar.

pom.xml
```diff
@@ -190,17 +190,67 @@
...
+                        <enabled>true</enabled>
+                    </snapshots>
+                </repository>
+            </repositories>
+        </profile>
+        <profile>
+            <id>public-repository</id>
+            <activation>
+                <property>
+                    <name>!useInternalRepo</name>
```
romanstrobl, line 237: I would prefer using: `<activeByDefault>true</activeByDefault>`.

pom.xml
```diff
@@ -190,17 +190,67 @@
                 <additionalparam>-Xdoclint:none</additionalparam>
             </properties>
         </profile>
+        <profile>
+            <id>internal-repository</id>
```
romanstrobl, line 194: I would prefer calling the profile `jfrog-repository` instead of `internal-repository` because the code is open source. Not a big deal, just something to consider.

romanstrobl (approved):
I am fine with the changes, I just added several suggestions for improved naming.

pom.xml
zcgandcomp: I actually rename it from jfrog-repository as we can migrate to internal repository of github instead of artifactory so I try to keep name agnostic

pom.xml
zcgandcomp: Not a maven expert, but I believe that activate by default is not correct as we do not want to active it if the property is used.  In paralel to other possible profiles - activeByDefault is activated only if no other profile is activated. And we will need multiple profiles = for signing e.g.

.github/workflows/maven-deploy.yml
zcgandcomp: so release_type - I suggest to avoid possible key words.

pom.xml
romanstrobl: Ok.

romanstrobl: Yes, you are right, the behaviour of `activeByDefault` is a bit confusing, sorry.

PR #452, opened 2023-01-21
Fix #451: Allow overriding length of decimalized signatures

### LINKED ISSUE #451: Allow overriding length of decimalized signatures
Currently, the length of a component is fixed for decimalized signatures. For some use cases, it would be nice to allow overriding the value (i.e., to achieve shorter offline code).

petrdvorak:


powerauth-docs/test-vectors/signatures-offline.json
```diff
@@ -0,0 +1,3124 @@
+{
+  "description" : "Client must be able to compute PowerAuth offline signature (using 1FA, 2FA signature keys) based on given data, counter and signature type",
+  "data" : [ {
+    "input" : {
+      "signaturePossessionKey" : "PTCBpEsGknUzlYG0Y9dumQ==",
+      "signatureKnowledgeKey" : "sGfvJAi5h6sxXV+vEG/2iQ==",
+      "signatureBiometryKey" : "otzFKCko5lcEGRev//uEbQ==",
+      "signatureType" : "possession",
+      "signatureComponentLength" : "2",
```
romanstrobl, line 9: I find this implementation misleading.

Why do we allow `signatureComponentLength` of `2`, when actual the signature length is `4`.

I think we should reject such configuration to avoid surprises in cryptography code and allow only lengths 4 - 12, as expected.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/config/SignatureConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class SignatureConfiguration {
+
+    /**
+     * Key used for retrieving expected length of the signature component.
+     */
+    public static final String DECIMAL_SIGNATURE_COMPONENT_LENGTH = "DECIMAL_SIGNATURE_COMPONENT_LENGTH";
+
+    private final PowerAuthSignatureFormat signatureFormat;
+    private final Map<String, Object> signatureParameters = new LinkedHashMap<>();
```
romanstrobl, line 37: Just wondering, why do we not use strong typing (e.g. enumeration) for signature parameters? The signature parameters are known and very specific for each release, right? Or is there some deeper reason behind using a hashmap?

romanstrobl (changes requested):
This is a nice improvement, however I do not like the "magical" behaviour. E.g. when signature length is set to 2, the actual signature length is 4. We should reject invalid setting in my opinion, because this can lead to misleading implementations.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -773,6 +777,95 @@ public void testSignatureValidationV31() throws Exception {
...
+    public void testSignatureOfflineValidation() throws Exception {
+
+        TestSet testSet = new TestSet("signatures-offline.json", "Client must be able to compute PowerAuth offline signature (using 1FA, 2FA signature keys) based on given data, counter and signature type");
+
+        int min = 2;
+        int max = 8;
+        int keyMax = 2;
+        int signatureCount = 10;
+        int dataMax = 256;
+        for (int j = min; j < max; j++) {
```
hvge, line 790: The configuration and loop above has several flaws:
- Doesn't test for default value equal to 8 (see your test vectors)
- Min value for "j" should start at 4. The actual implementation reject minimal length equal to 4, so we should generate test data from a valid test input.

powerauth-docs/test-vectors/signatures-offline.json
hvge: I found the same issue. Also there's a problem that the test generator doesn't generate vectors for the default signatureComponentLength equal to 8

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/config/SignatureConfiguration.java
hvge: I agree with this, the same issue for me.

hvge (changes requested):
I found mostly the same issues as @romanstrobl

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
romanstrobl: How does the max value of 8 (exclusive) relate to:

```
int signatureDecimalLength = (length == null) ? PowerAuthConfiguration.SIGNATURE_DECIMAL_LENGTH : Math.min(Math.max(length, 4), 12);
```

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -98,21 +98,27 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
         final List<byte[]> signatureComponents = computePowerAuthSignatureComponents(data, signatureKeys, ctrData);
+        // Determine the length of the signature component
+        int signatureDecimalLength = (length == null) ? PowerAuthConfiguration.SIGNATURE_DECIMAL_LENGTH : Math.min(Math.max(length, 4), 12);
         // Convert byte components into decimal signature
         for (int i = 0; i < signatureComponents.size(); i++) {
             final byte[] signatureComponent = signatureComponents.get(i);
             int index = signatureComponent.length - 4;
-            int number = (ByteBuffer.wrap(signatureComponent).getInt(index) & 0x7FFFFFFF) % (int) (Math.pow(10, PowerAuthConfiguration.SIGNATURE_DECIMAL_LENGTH));
-            signatureStringComponents[i] = String.format("%0" + PowerAuthConfiguration.SIGNATURE_DECIMAL_LENGTH + "d", number);
+            int number = (ByteBuffer.wrap(signatureComponent).getInt(index) & 0x7FFFFFFF) % (int) (Math.pow(10, signatureDecimalLength));
```
hvge, line 120: Several lines above you set the upper limit to 12 which will cause integer to overflow here. 10^12 is way bigger than 2^31.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
```diff
@@ -98,21 +98,27 @@ public boolean validateECDSASignature(byte[] signedBytes, byte[] signature, Publ
...
      * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
      */
-    private String computePowerAuthDecimalSignature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData) throws GenericCryptoException, CryptoProviderException {
+    private String computePowerAuthDecimalSignature(byte[] data, List<SecretKey> signatureKeys, byte[] ctrData, Integer length) throws GenericCryptoException, CryptoProviderException {
         // Prepare holder for signature components
         final String[] signatureStringComponents = new String[signatureKeys.size()];
         // Compute signature components
         final List<byte[]> signatureComponents = computePowerAuthSignatureComponents(data, signatureKeys, ctrData);
+        // Determine the length of the signature component
+        int signatureDecimalLength = (length == null) ? PowerAuthConfiguration.SIGNATURE_DECIMAL_LENGTH : Math.min(Math.max(length, 4), 12);
```
hvge, line 115: Upper limit should be 8

hvge (changes requested):

powerauth-docs/test-vectors/signatures-offline.json
petrdvorak: @hvge @romanstrobl The conversion methods currently do not throw any exceptions or return error codes. As a result, we might break the behavior (what happens if we randomly throw an exception somewhere before computing the signature?). Submitting the length 2, 3, -1 and obtaining length 4 is legitimate behavior we can proxy on other levels (i.e., in PowerAuth Server or PowerAuth Cloud).

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/config/SignatureConfiguration.java
petrdvorak: @hvge @romanstrobl Yes, there is a deeper reason. The type of signature is dynamically determined based on a signature header. We need to adjust the parameters later. Sure, if you can enumerate all possible use cases we will encounter in the next 5-10 releases, we can strongly type it.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
petrdvorak: @romanstrobl It does not. Why should it?
@hvge This is not a test, we generate test vectors here.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/SignatureUtils.java
petrdvorak: Good point.

petrdvorak:
I will have a look.

powerauth-docs/test-vectors/signatures-offline.json
petrdvorak: Fixed.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/config/SignatureConfiguration.java
petrdvorak: Fixed.

hvge (approved):
Looks OK now. The last thing that's missing is a short note in the documentation about this option.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/config/SignatureConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+    public static SignatureConfiguration forFormat(PowerAuthSignatureFormat format) throws CryptoProviderException {
+        switch (format) {
+            case BASE64: {
+                return new Base64SignatureConfiguration();
+            }
+            case DECIMAL: {
+                return new DecimalSignatureConfiguration();
+            }
+        }
+        throw new CryptoProviderException("Invalid or null format provided");
```
banterCZ, line 57: May we include `format` value itself in the exception message?

petrdvorak: OK

romanstrobl (approved):
Nice solution for the configuration, thanks!

PR #458, opened 2023-03-02
Fix for KEY_MASTER_SECRET documentation

hvge:
This PR fixes documentation of KEY_MASTER_SECRET calculation.


romanstrobl (approved):
Good catch!

PR #463, opened 2023-04-04
Docs: Fixed KDF definition and usage

hvge:
This PR fixes KDF function usage and the definition in the documentation.


petrdvorak (approved):
Looks good

PR #465, opened 2023-04-10
Fix #464: Create a method to construct a public key based on X and Y coordinates

### LINKED ISSUE #464: Create a method to construct a public key based on X and Y coordinates

petrdvorak:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -103,6 +103,49 @@ public PublicKey convertBytesToPublicKey(byte[] keyBytes) throws InvalidKeySpecE
...
+     * @return Public key with provided coordinates.
+     * @throws InvalidKeySpecException When provided bytes are not a correct key representation.
+     * @throws CryptoProviderException When crypto provider is incorrectly initialized.
+     * @throws GenericCryptoException  When public key is invalid.
+     */
+    public PublicKey convertPointBytesToPublicKey(byte[] xBytes, byte[] yBytes) throws InvalidKeySpecException, CryptoProviderException, GenericCryptoException {
+        try {
+            // Make sure the values are interpreted as positive integer.
+            final BigInteger x = new BigInteger(1, xBytes);
+            final BigInteger y = new BigInteger(1, yBytes);
```
romanstrobl, line 120: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -103,6 +103,49 @@ public PublicKey convertBytesToPublicKey(byte[] keyBytes) throws InvalidKeySpecE
...
+
+            // Validate the point is correct
+            final ECNamedCurveParameterSpec ecSpec = ECNamedCurveTable.getParameterSpec("secp256r1");
+            if (ecSpec == null) { // can happen with incorrectly initialized crypto provider.
+                throw new CryptoProviderException("Crypto provider does not support the secp256r1 curve");
+            }
+            final org.bouncycastle.math.ec.ECPoint point = ecSpec.getCurve().createPoint(x, y);
+            publicKeyValidator.validate(ecSpec.getCurve(), point);
+
+            final BigInteger xAffine = point.getAffineXCoord().toBigInteger();
```
romanstrobl, line 130: We already know this value, why do we need to extract it?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -103,6 +103,49 @@ public PublicKey convertBytesToPublicKey(byte[] keyBytes) throws InvalidKeySpecE
...
+            final BigInteger yAffine = point.getAffineYCoord().toBigInteger();
```
romanstrobl, line 131: dtto

romanstrobl (approved):
This is a nice addition, however I have two ideas which could improvecode quality to consider.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -103,6 +103,49 @@ public PublicKey convertBytesToPublicKey(byte[] keyBytes) throws InvalidKeySpecE
...
+     * @param yBytes Y coordinate.
+     * @return Public key with provided coordinates.
+     * @throws InvalidKeySpecException When provided bytes are not a correct key representation.
+     * @throws CryptoProviderException When crypto provider is incorrectly initialized.
+     * @throws GenericCryptoException  When public key is invalid.
+     */
+    public PublicKey convertPointBytesToPublicKey(byte[] xBytes, byte[] yBytes) throws InvalidKeySpecException, CryptoProviderException, GenericCryptoException {
+        try {
+            // Make sure the values are interpreted as positive integer.
+            final BigInteger x = new BigInteger(1, xBytes);
```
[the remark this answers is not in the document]
petrdvorak: This does not work like this - you typically get `byte[]` that can be this or that, and you need to tell that you want to use the bytes as positive `BigInteger`...

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
petrdvorak: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java
petrdvorak: dtto

romanstrobl:
Note that this method is not consistent with the other public key import method in powerauth-crypto, which uses byte array in ASN.1 format, which is parsed by the Bouncy Castle library and the point is validated:
https://github.com/wultra/powerauth-crypto/blob/e334f81df7872afcb2ec0552e520e9775bb3ed2f/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/KeyConvertor.java#L78

PR #467, opened 2023-04-12
Fix #466: Fix public key conversion from point

### LINKED ISSUE #466: Fix public key conversion from point
We simplified the public key conversion logic a bit too much.

petrdvorak:


romanstrobl (approved):
Alright, let's do the conversion this way.

PR #472, opened 2023-04-26
Fix #471: Add support for Time-Based One-Time Password (TOTP)

### LINKED ISSUE #471: Add support for Time-Based One-Time Password (TOTP)
https://www.rfc-editor.org/rfc/rfc6238

banterCZ:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+    /**
+     * Generates a TOTP value for the given set of parameters using HmacSHA1 algorithm.
+     *
+     * @param key           the shared secret
+     * @param localDateTime date time
+     * @param returnDigits  number of digits to return
+     * @return a numeric String in base 10 that includes truncation digits
+     * @throws CryptoProviderException in case of any crypto error
+     */
+    public static String generateTOTPSHA1(final byte[] key, final LocalDateTime localDateTime, final int returnDigits) throws CryptoProviderException {
```
romanstrobl, line 59: This camel case is not consistent with how we usually use camel case, I would write `generateTotpSha1`.

romanstrobl: Why do we allow usage of `SHA1`? This algorithm is vulnerable, is this option required?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+        if (steps < 0) {
+            throw new CryptoProviderException("Steps must not be negative number");
+        }
+
+        long currentTimeStep = countTimeSteps(localDateTime);
+        for (int i = 0; i <= steps; i++) {
+            logger.debug("Validating TOTP for localDateTime={}, algorithm={}, backward step={} out of allowed steps={}", localDateTime, algorithm, i, steps);
+            final long step = currentTimeStep - i;
+            final String expectedOtp = generateTOTP(key, step, otp.length(), algorithm);
+            if (expectedOtp.equals(otp)) {
```
romanstrobl, line 200: This section is prone to side channel timing attacks, although the TOTP is really short, we should probable still avoid code branches. We should compare all parts of the TOTP everytime.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.util.HexFormat;
+
+/**
+ * TOTP: Time-Based One-Time Password Algorithm according to <a href="https://www.rfc-editor.org/rfc/rfc6238">RFC 6238</a>.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public final class TOTP {
```
romanstrobl, line 36: I do not like the allcaps name of this class. Can we improve this?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(TOTP.class);
+
+    private static final int TIME_STEP_X = 30;
```
romanstrobl, line 40: Can we make this option configurable? It looks like a magic constant in code.

romanstrobl (changes requested):
In general, it should work fine, however I found several issues which I believe should be addressed.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+        if (returnDigits <= 0) {
+            throw new CryptoProviderException("ReturnDigits must be positive number");
+        }
+
+        // Using the counter
+        // First 8 bytes are for the movingFactor
+        // Compliant with base RFC4226 (HOTP)
+        final String timeWithCounterPrefix = padWithZeros(Long.toHexString(timeStep), 16);
+
+        // Get the HEX in a Byte[]
```
romanstrobl, line 238: The comment should probably say `byte[]` instead of `Byte[]`, a bit misleading.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
romanstrobl, line 200: Please consider using `Arrays.constantTimeAreEqual` from Bouncy Castle to avoid any (even potential) timing attacks. These constant time comparions are usually done over `byte[]`.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+        final int offset = hash[hash.length - 1] & 0xf;
+
+        final int binary = ((hash[offset] & 0x7f) << 24) |
+                ((hash[offset + 1] & 0xff) << 16) |
+                ((hash[offset + 2] & 0xff) << 8) |
+                (hash[offset + 3] & 0xff);
+
+        final int otp = binary % DIGITS_POWER[returnDigits];
+
+        final String result = Integer.toString(otp);
```
romanstrobl, line 253: Is there a reason for using String here? I would prefer using `byte[]` for such operations and only convert the result to String, if required.

powerauth-java-crypto/src/test/resources/io/getlime/security/powerauth/crypto/lib/totp/data.csv
```diff
@@ -0,0 +1,19 @@
+ Time (sec) ,  UTC Time           , Value of T (hex) ,  TOTP    , Algorithm
```
romanstrobl, line 1: It seems that we modified the test vectors from the RFC. Can we stick to the test vectors from the RFC?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
banterCZ: We may configure it if needed, but is that comment enough?

> We RECOMMEND a default time-step size of 30 seconds.  This default
   value of 30 seconds is selected as a balance between security and
   usability.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/totp/TOTPTest.java
```diff
@@ -0,0 +1,110 @@
...
+    static class DateTimeConverter extends TypedArgumentConverter<String, LocalDateTime> {
+        protected DateTimeConverter() {
+            super(String.class, LocalDateTime.class);
+        }
+
+        @Override
+        protected LocalDateTime convert(final String source) throws ArgumentConversionException {
+            try {
+                return LocalDateTime.parse(source);
+            } catch (Exception e) {
```
romanstrobl, line 105: Can we add a debug log for this exception before converting it?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+     * @param text      the message or text to be authenticated
+     * @throws CryptoProviderException in case of any crypto error
+     */
+    private static byte[] computeHash(final String algorithm, final byte[] keyBytes, final byte[] text) throws CryptoProviderException {
+        try {
+            final Mac hmac = Mac.getInstance(algorithm);
+            final SecretKeySpec macKey = new SecretKeySpec(keyBytes, "RAW");
+            hmac.init(macKey);
+            return hmac.doFinal(text);
+        } catch (GeneralSecurityException e) {
```
romanstrobl, line 283: Can we add a debug log for this exception before converting it?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+
+    /**
+     * Computes a Hashed Message Authentication Code with the give hash algorithm as a parameter.
+     *
+     * @param algorithm the algorithm
+     * @param keyBytes  the bytes to use for the HMAC key
+     * @param text      the message or text to be authenticated
+     * @throws CryptoProviderException in case of any crypto error
+     */
+    private static byte[] computeHash(final String algorithm, final byte[] keyBytes, final byte[] text) throws CryptoProviderException {
```
romanstrobl, line 277: The variable name `text` is a bit confusing, because it is of `byte[]` type. I would use something like `input` or `data`, not `text`.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+
+        // Get the HEX in a Byte[]
+        final byte[] msg = HexFormat.of().parseHex(timeWithCounterPrefix);
+
+        final byte[] hash = computeHash(algorithm, key, msg);
+
+        // put selected bytes into result int
+        final int offset = hash[hash.length - 1] & 0xf;
+
+        final int binary = ((hash[offset] & 0x7f) << 24) |
```
romanstrobl, line 246: I am a bit confused about the variable name here: `int binary`. Can we improve the name a bit so that it corresponds to the step?

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/lib/totp/TOTPTest.java
banterCZ: I believe that it is not needed. This is a test converter and it is clearly propagated to the test report.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
banterCZ: Added logger.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
banterCZ: Renamed and sha1 dropped.

powerauth-java-crypto/src/test/resources/io/getlime/security/powerauth/crypto/lib/totp/data.csv
banterCZ: Fixed, in RFC there is a wrong comment about seed for the test data.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
```diff
@@ -0,0 +1,299 @@
...
+        if (otp == null || otp.trim().length() == 0) {
+            throw new CryptoProviderException("Otp is mandatory");
+        }
+
+        if (steps < 0) {
+            throw new CryptoProviderException("Steps must not be negative number");
+        }
+
+        long currentTimeStep = countTimeSteps(localDateTime);
+        for (int i = 0; i <= steps; i++) {
```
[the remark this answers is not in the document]
banterCZ: Yes, I am. I will rename it to `backwardSteps`, covered by test cases. This statement must run even for zero.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/TOTP.java
banterCZ: Do you have an idea to improve? I took that from RFC.

romanstrobl: I would use `binaryCode`, this corresponds to the description of original HOTP algorithm in [RFC 4226](https://www.rfc-editor.org/rfc/rfc4226).

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/Totp.java
```diff
@@ -0,0 +1,258 @@
...
+
+        if (otp == null) {
+            throw new CryptoProviderException("Otp is mandatory");
+        }
+
+        if (backwardSteps < 0) {
+            throw new CryptoProviderException("Steps must not be negative number");
+        }
+
+        long currentTimeStep = countTimeSteps(localDateTime);
```
romanstrobl, line 158: Variable can be final.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/totp/Totp.java
```diff
@@ -0,0 +1,258 @@
...
+        // Using the counter
+        // First 8 bytes are for the movingFactor
+        // Compliant with base RFC4226 (HOTP)
+        final byte[] data = HexFormat.of().parseHex(padWithZeros(Long.toHexString(timeStep), 16));
+        final byte[] hash = computeHash(algorithm, key, data);
+
+        // Put selected bytes into result int
+        final int offset = hash[hash.length - 1] & 0xf;
+
+        final int binaryCode = ((hash[offset] & 0x7f) << 24) |
```
romanstrobl, line 205: Can we add some comment about the `0x7f` value?

romanstrobl (approved):
I have no additional comments, good job!

hvge (changes requested):
The implementation looks OK, but I have a problem with the validation functions.

I think that we should be more more paranoid here and add an expected length of OTP as input of the validation. The problem now is that if attacker controls inputs and can alter OTP, then he can simply use one character length and try all 10 possibilities.

In the current implementation we expect that caller is responsible for such validation. This is not a good practice in cryptography :)

Otherwise good work.

petrdvorak (approved):

PR #490, opened 2023-07-19
Fix #489: Make possible to configure time step length of TOTP

### LINKED ISSUE #489: Make possible to configure time step length of TOTP
Right now, the recommended value of 30 seconds is used, but for some use cases, it should be possible to override it.

A follow-up to #471

banterCZ:


romanstrobl (approved):
Looks OK.

PR #495, opened 2023-08-04
Fix 494: Remove remaining code related to V2 protocol

hvge:
This PR removes support for protocol V2 from crypto library


powerauth-java-http/src/main/java/io/getlime/security/powerauth/http/validator/ValueTypeValidator.java
```diff
@@ -52,7 +52,7 @@ public class ValueTypeValidator {
      * Admissible protocol versions in the header.
      */
     private static final Set<String> PROTOCOL_VERSIONS = new HashSet<>(Arrays.asList(
-            "3.1", "3.0", "2.1", "2.0"
+            "3.1", "3.0"
```
banterCZ, line 55: Nitpick: If we touch this line, we may use JDK 17+, making it also immutable.
```suggestion
private static final Set<String> PROTOCOL_VERSIONS = Set.of("3.1", "3.0");
```

banterCZ (approved):
Looks good.

Issue #492, opened 2023-08-04
Crypto 3.2: Document changes in protocol


hvge:
Update public documentation to match our internal specification for Crypto 3.2 protocol.

PR #486, opened 2023-05-31
Fix #485: Crypto 3.2: Prevent replay attacks

romanstrobl:

PR #497, opened 2023-08-09
Simplified End-2-End Encryption

hvge:
General changes:

- Added high level `EncryptorFactory` that provide encryptors both for server and the client side for all protocol versions.
- Added `ClientEncryptor` and `ServerEncryptor` interfaces providing client or server-side encryption.
- Each implementation holds the context for the next opposite operation. For example, `ClientEncryptor` allow us to implement one (or more) round-trip when the request is encrypted and the appropriate response from the server is decrypted. The similar principle is valid for the server side.
- Each type of encryptor is parametrized with common `EncryptorParameters` and custom `EncryptorSecrets` where secrets depends on actual implementation and the side of the encryption.
- We suppose to do not use low level `EciesEncryptor` and `EciesDecryptor` anymore in other projects.

Refactoring:
- `EciesScope` is now `EncryptorScope`
- `EciesSharedInfo1` is now `EncryptorId`. The `EncryptorId` also contains information about ECIES SharedInfo1 and the scope. This small change simplified a lot of things on behalf.

Additional fixes:
- #498
- #492


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/ClientEciesEncryptor.java
```diff
@@ -0,0 +1,221 @@
...
+    private final EciesRequestResponseValidator validator;
+    private final byte[] associatedData;        // non-null for V3.2+
+
+    // Variables altered after configureKeys() call.
+    private ClientEncryptorSecrets encryptorSecrets;
+
+    /**
+     * Get SharedInfo2 base bytes. The function should be used only for the testing purposes.
+     */
+    @Getter
```
petrdvorak, line 59: Not sure if this light usage of Lombok is necessary.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/EciesUtils.java
```diff
@@ -54,22 +60,96 @@ public static long generateTimestamp() {
...
      * @return Derived associated data.
+     * @throws EciesException In case that activation ID is required but is missing.
      */
-    public static byte[] deriveAssociatedData(EciesScope eciesScope, String version, String applicationKey, String activationId) {
-        if ("3.2".equals(version)) {
-            if (eciesScope == EciesScope.ACTIVATION_SCOPE) {
-                return ByteUtils.concatStrings(version, applicationKey, activationId);
+    public static byte[] deriveAssociatedData(EncryptorScope scope, String protocolVersion, String applicationKey, String activationId) throws EciesException {
+        if ("3.2".equals(protocolVersion)) {
+            if (protocolVersion == null) {
```
petrdvorak, line 72: This can never happen, the `protocolPersion` variable is `"3.3"` according to the condition just above... the condition should probably be above?

petrdvorak (changes requested):
There is one fishy condition to review. The use of Lombok is OK. In some classes, however, it was so minimal that I wonder if adding an explicit getter isn't a better choice (especially on classes not annotated with `@Data` annotation).

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/ClientEciesEncryptor.java
hvge: Good catch. At some point I needed the values in tests, but then I changed how we test the encryption in general.

petrdvorak (approved):
Looks OK

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/EncryptorFactory.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * The {@code EncryptorFactory} class provide high level encryptors for PowerAuth End-To-End encryption implementation.
+ * You can construct both server and client-side constructors in this factory.
+ */
+public class EncryptorFactory {
+    /**
+     * Default constructor.
+     */
+    public EncryptorFactory() {}
```
banterCZ, line 37: Nitpick: It is not default, but no-arg constructor and is not needed.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java
```diff
@@ -0,0 +1,75 @@
...
+        if (request.getEphemeralPublicKey() == null || request.getEncryptedData() == null || request.getMac() == null) {
+            return false;
+        }
+        if (useNonceForRequest == (request.getNonce() == null)) {
+            // Fails when nonce is missing in 3.1+
+            // Fails when nonce is present in 3.0
+            return false;
+        }
+        return useTimestamp == (request.getTimestamp() != null);
+        // Fails when timestamp is missing in 3.2+
```
banterCZ, line 53: 👀 Nitpick: I am surprised by a comment bellow the line.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java
```diff
@@ -0,0 +1,75 @@
...
+        if (useTimestamp) {
+            // 3.2+
+            return response.getNonce() != null && response.getTimestamp() != null;
+        } else {
+            // 3.0, 3.1 should not contain nonce and timestamp in response
+            return response.getNonce() == null && response.getTimestamp() == null;
+        }
+    }
+
+    private final static Set<String> supportedVersions = Set.of("3.2", "3.1", "3.0");
```
banterCZ, line 74: Nitpick: In Java world, I would expect this at the very beginning.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/EncryptedRequest.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * The {@code EncryptedRequest} object represents an encrypted request payload in PowerAuth End-To-End encryption scheme.
+ */
+@Data
+@AllArgsConstructor
+public class EncryptedRequest {
+    private String ephemeralPublicKey;
+    private String encryptedData;
+    private String mac;
+    private String nonce;
```
banterCZ, line 32: Should be `nonce` part of `toString`?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/EncryptedResponse.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The {@code EncryptedResponse} object represents an encrypted request payload in PowerAuth End-To-End encryption scheme.
+ */
+@Data
+@AllArgsConstructor
+public class EncryptedResponse {
+    private String encryptedData;
+    private String mac;
+    private String nonce;
```
banterCZ, line 31: dtto

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/v3/ClientEncryptorSecrets.java
```diff
@@ -0,0 +1,98 @@
...
+@Data
+public class ClientEncryptorSecrets implements EncryptorSecrets {
+    /**
+     * Server's public key.
+     */
+    private final PublicKey serverPublicKey;
+    /**
+     * Application secret.
+     */
+    private final String applicationSecret;
```
banterCZ, line 45: dtto: toString

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/v3/ServerEncryptorSecrets.java
```diff
@@ -0,0 +1,105 @@
...
+@Data
+public class ServerEncryptorSecrets implements EncryptorSecrets {
+    /**
+     * Server's private key.
+     */
+    private final PrivateKey serverPrivateKey;
+    /**
+     * Application's secret.
+     */
+    private final String applicationSecret;
```
banterCZ, line 45: dtto: toString

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/encryption/GeneralEncryptorTest.java
```diff
@@ -0,0 +1,959 @@
...
+            EncryptorId.APPLICATION_SCOPE_GENERIC,
+            EncryptorId.ACTIVATION_SCOPE_GENERIC,
+            EncryptorId.ACTIVATION_LAYER_2,
+            EncryptorId.CREATE_TOKEN,
+            EncryptorId.VAULT_UNLOCK,
+            EncryptorId.UPGRADE,
+            EncryptorId.CONFIRM_RECOVERY_CODE
+    );
+
+    private String APPLICATION_KEY;
```
banterCZ, line 69: It does not comply with Java conventions. It should be either lower-case or _final static_.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/EncryptorFactory.java
hvge: Removed

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/v3/ClientEncryptorSecrets.java
hvge: Such secret is not so sensitive, but other properties in the class are. I'll exclude them all.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/model/v3/ServerEncryptorSecrets.java
hvge: Such secret is not so sensitive, but other properties in the class are. I'll exclude them all.

powerauth-java-crypto/src/test/java/io/getlime/security/powerauth/crypto/encryption/GeneralEncryptorTest.java
hvge: I got inspiration from original ECIES test. Such values are typically constants hardcoded in the mobile app (e.g. final static), and that's the reason why it break rules. From the test-case perspective, the value doesn't change during the one particular test. We don't want then static and final, because it's good to use a different values for each test.

romanstrobl (approved):
Very nice cleanup of the code and improved support for cryptography setup in server components. I only have a small comment related to different naming in documentation and implementation in `ByteUtils` class.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/ClientEciesEncryptor.java
```diff
@@ -0,0 +1,218 @@
...
+                sharedInfo2Base,
+                null,
+                responseNonce,
+                responseTimestamp,
+                associatedData
+        );
+        // Build decryptor object.
+        final EciesDecryptor decryptor = new EciesDecryptor(envelopeKey, sharedInfo2);
+        // Prepare EciesPayload and try to decrypt data.
+        final byte[] plaintext = decryptor.decrypt(new EciesPayload(
```
romanstrobl, line 195: This code is hard to read, can we please separate this code into multiple lines and introduce 2 variables for new objects?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/ServerEciesEncryptor.java
```diff
@@ -0,0 +1,265 @@
...
+                sharedInfo2Base,
+                ephemeralPublicKey,
+                requestNonce,
+                requestTimestamp,
+                associatedData
+        );
+        // Once we have SharedInfo2 prepared, we can construct a decryptor.
+        final EciesDecryptor eciesDecryptor = new EciesDecryptor(envelopeKey, sharedInfo2);
+        // Prepare EciesPayload and try to decrypt data.
+        final byte[] plaintext = eciesDecryptor.decrypt(new EciesPayload(
```
romanstrobl, line 188: This code is hard to read, can we please separate this code into multiple lines and introduce 2 variables for new objects?

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ClientEncryptor.java
```diff
@@ -0,0 +1,75 @@
...
+     * Get this encryptor's identifier.
+     * @return This encryptor's identifier.
+     */
+    EncryptorId getEncryptorId();
+
+    /**
+     * Get parameters used to construct this encryptor.
+     * @return EncryptorParameters used to construct this encryptor.
+     */
+    EncryptorParameters getParameters();
```
romanstrobl, line 38: `ServerEncryptor` uses `getEncryptorParameters` method name, perhaps we should unify these method names. The order of methods is also a litle bit different between these two interfaces.

docs/Basic-definitions.md
```diff
@@ -273,28 +273,96 @@ Get last N bytes of given byte array.
...
 long value = ByteUtils.getLong(byte[8] bytes);
 ```
 
-#### Concatenate Arrays
+#### Encode Primitive Types To Byte Array
+
+Encode short value into byte array in big endian order.
+
+```java
+byte[] encoded = ByteUtils.encode(short n);
```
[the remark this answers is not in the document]
hvge: I wonder why you don't use method overloading in java :/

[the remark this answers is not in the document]
romanstrobl: See previous PR - the problem is, that it is easy to use a wrong method because of automated type conversions in Java.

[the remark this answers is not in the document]
romanstrobl: The previous discussion: https://github.com/wultra/powerauth-crypto/pull/486#discussion_r1214420195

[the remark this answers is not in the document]
hvge: I would keep it as it is. This is a specification written in pseudocode (annotated as java, for the syntax highlighting). From the context it's clear what's going on. I would prefer simpler documentation in this case.

PR #519, opened 2023-08-31
Fix #518: Dereference null return value warning

### LINKED ISSUE #518: Dereference null return value warning
Coverity scan detected possible null pointer exception:

```
CID 399720 (#1 of 1): Dereference null return value (NULL_RETURNS)
6. dereference: Dereferencing a pointer that might be null requestNonce when calling encodeToString.
```

The issue was detected in class `ClientEciesEncryptor`.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #525, opened 2023-09-06
Fix #524: Include version in the token digest

### LINKED ISSUE #524: Include version in the token digest
Currently, we compute token digest based on timestamp and nonce, we should add version too.

petrdvorak:


romanstrobl (approved):
Looks OK.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/TokenUtils.java
```diff
@@ -92,29 +92,33 @@ public byte[] convertTokenTimestamp(long timestamp) {
...
      * @throws CryptoProviderException In case cryptography provider is incorrectly initialized.
      */
-    public byte[] computeTokenDigest(byte[] nonce, byte[] timestamp, byte[] tokenSecret) throws GenericCryptoException, CryptoProviderException {
-        byte[] amp = "&".getBytes(StandardCharsets.UTF_8);
-        byte[] data = ByteUtils.concat(nonce, amp, timestamp);
+    public byte[] computeTokenDigest(byte[] nonce, byte[] timestamp, byte[] version, byte[] tokenSecret) throws GenericCryptoException, CryptoProviderException {
+        final byte[] amp = "&".getBytes(StandardCharsets.UTF_8);
+        final byte[] data = (version != null)
+                ? ByteUtils.concat(nonce, amp, timestamp, amp, version)
+                : ByteUtils.concat(nonce, amp, timestamp);
```
hvge, line 105: I think that we should make version mandatory parameter and the logic whether we include it to calculation or not should be implemented here. Like  `if ("3.2".equals(version)) {...}`. Otherwise it will be yet another example where we delegate the protocol related logic to outside of the library. The result is a lot more spaghetti copy-pasted code in three (or more) repositories.

hvge (changes requested):

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/util/TokenUtils.java
hvge: At least we should do this logic in `ClientTokenGenerator` or `ServerTokenVerifier` classes.

PR #527, opened 2023-09-07
Fix #526: Fix incorrect error codes for the header validators

### LINKED ISSUE #526: Fix incorrect error codes for the header validators
The error codes for version validations are incorrectly named, this was not issue but now we start deprecating protocol versions.

petrdvorak:


romanstrobl (approved):
Looks OK.

hvge (approved):

PR #530, opened 2023-09-07
Fix #529: Fix typo in the error message

### LINKED ISSUE #529: Fix typo in the error message

petrdvorak:
Optimize imports


romanstrobl (approved):
Good catch!

banterCZ (approved):
A nice catch but but pity about the organizing imports in the not affected classes. It was not needed and in general, it complicates backporting and cherry-picking.

hvge (approved):

PR #549, opened 2023-12-03
Fix #489: Make possible to configure time step length of TOTP (backport)

### LINKED ISSUE #489: Make possible to configure time step length of TOTP

jandusil:
cherry picked from commit 4e4f6dde9c35a5d4a18621835b7d2ec7c3f0291d


jandusil:
Waiting for https://github.com/wultra/powerauth-crypto/pull/549

PR #569, opened 2023-12-20
Fix #554: Remove Guava dependency

### LINKED ISSUE #554: Remove Guava dependency
We should be able to do without it. The usage of `Base32` could be replaced by bouncycastle.
Let's target this into version 1.7.0

banterCZ:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
```diff
@@ -56,6 +56,11 @@ public class IdentifierGenerator {
      */
     private static final int ACTIVATION_CODE_BYTES_LENGTH = 12;
 
+    /**
+     * When {@code ACTIVATION_CODE_BYTES_LENGTH = 12}, the Base32 padding is always {@code ====}.
+     */
+    private static final String ACTIVATION_CODE_PADDING = "====";
```
romanstrobl, line 62: What happens in case `ACTIVATION_CODE_BYTES_LENGTH` constant would change? Isn't it better to calculate the padding characters?

romanstrobl (approved):
I have one minor concern about the padding.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
banterCZ: 🤔 I am not merging it yet. I will return to this later. The thing is that BC decoder is much more strict and unlike Guva it throws an exception if it is without padding. So we transform for example
`AAAAAAAAAAAAAAAAAAAA====` into `AAAAA-AAAAA-AAAAA-AAAAA` but decoding of `AAAAAAAAAAAAAAAAAAAA` would fail.

romanstrobl: See: https://github.com/bcgit/bc-java/blob/4ee2ee57b2ed6e8a80120c5c90796438701e2b88/core/src/main/java/org/bouncycastle/util/encoders/Base32Encoder.java#L90

banterCZ: Thanks for the hint. Now the padding is not fixed anymore but calculated.

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/generator/IdentifierGenerator.java
```diff
@@ -184,6 +179,32 @@ public boolean validateActivationCode(String activationCode) {
...
+     * @return base32 with padding
+     */
+    private static String fetchActivationCodeBase32(final String activationCode) {
+        final String activationCodeWithoutHyphens = activationCode.replace("-", "");
+        // The activation code does not contain the padding, but it must be present in the Base32 value to be valid.
+        final String activationCodePadding = switch (activationCodeWithoutHyphens.length() % 8) {
+            case 2:
+                yield "======";
+            case 4:
+                yield  "====";
```
romanstrobl, line 197: Please fix the formatting (double space).

romanstrobl (approved):
Thanks for the padding calculation.

PR #606, opened 2024-04-22
Fix #605: Refactor BouncyCastle dependency

### LINKED ISSUE #605: Refactor BouncyCastle dependency
Let's change the scope of BC dependency from `provided` to `compile`.

banterCZ:
- Make the scope `compile` instead of `provided`

### DIFF
--- a/powerauth-java-crypto/pom.xml
+++ b/powerauth-java-crypto/pom.xml
@@ -60,7 +60,6 @@
 			<groupId>org.bouncycastle</groupId>
 			<artifactId>bcprov-jdk18on</artifactId>
 			<version>1.78.1</version>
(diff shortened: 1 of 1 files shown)


romanstrobl (approved):
Please test this change with all application servers, especially JBoss/Wildfly and Weblogic.

banterCZ:
> Please test this change with all application servers, especially JBoss/Wildfly and Weblogic.

This change does not affect JBoss itself. I will test it in PAS https://github.com/wultra/powerauth-server/issues/1476 
WebLogic does not yet support EE 9.1 or 10 anyway https://en.wikipedia.org/wiki/Oracle_WebLogic_Server#Standards_support_by_version

PR #617, opened 2024-06-07
Fix #616: Add a note on built-in keys to the list of used keys

### LINKED ISSUE #616: Add a note on built-in keys to the list of used keys
We currently document derived keys, we should probably also add section for built-in keys (even though these are documented in the docs for the process).

petrdvorak:


docs/List-of-used-keys.md
```diff
@@ -2,14 +2,23 @@
 
 The following keys are used in the PowerAuth cryptography scheme.
 
+## Application Scoped Keys
+
+| name                        | created as                               | purpose                                                                                                                                                             |
```
banterCZ, line 7: This is not rendered as a table.

banterCZ (approved):

PR #641, opened 2024-09-06
Fix #640: Add documentation for temporary keys

### LINKED ISSUE #640: Add documentation for temporary keys
We implemented a new feature with temporary encryption keys that should be documented in the scope of the base crypto protocol.

petrdvorak:


hvge (changes requested):
I think we should also update a [End-To-End-Encryption](https://developers.wultra.com/components/powerauth-crypto/1.8.x/documentation/End-To-End-Encryption) section. There's still information that ECDH is computed with static keys. Also there's missing information about `temporaryKeyId` attribute in the payload.

docs/End-To-End-Encryption.md
```diff
@@ -56,17 +59,17 @@ Note that the `APPLICATION_SECRET` constant is in Base64 form, so we need to rei
 
 #### Activation Scope
 
-ECIES in activation scope has following configuration of parameters:
+ECIES in activation scope has the following configuration of parameters:
 
-- `KEY_ENC_PUB` is `KEY_SERVER_PUBLIC` (e.g. key which is unique for each activation)
+- `KEY_ENC_PUB` is a [temporary key](./Temporary-Encryption-Keys.md) with given `TEMP_KEY_ID` identifier fetched from the server associated with a specific application version and activation, and signed with `KEY_SERVER_PUBLIC` (the key which is unique for each activation, to prove it was intended for the activations cope).
```
hvge, line 64: Signed with `KEY_SERVER_PRIVATE`.

hvge (changes requested):
Just one small fix and it would be OK.

docs/End-To-End-Encryption.md
petrdvorak: Thank you!

PR #643, opened 2024-09-17
Fix #642: Add exact list of used cryptographic standards

### LINKED ISSUE #642: Add exact list of used cryptographic standards
To simplify reviews of the algorithm, we should add a list of algorithms used in the protocol. While this is generally useless information, some auditors do not look at the big picture and just skim the algorithms, so we need to be ready for those...

petrdvorak:


zcgandcomp (approved):
Nice

PR #655, opened 2024-10-03
Fix #654: Add temporary keys into the list of keys

### LINKED ISSUE #654: Add temporary keys into the list of keys
The temporary keys used in ECIES should be added to the list of keys.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #664, opened 2024-10-14
Fix #663: Enforce check for temporary keys in protocol version 3.3

### LINKED ISSUE #663: Enforce check for temporary keys in protocol version 3.3
The temporary key check is missing in validator for protocol version 3.3:

https://github.com/wultra/powerauth-crypto/blob/develop/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java

romanstrobl:


powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java
```diff
@@ -65,6 +70,9 @@ public boolean validateEncryptedRequest(EncryptedRequest request) {
         if (request.getEphemeralPublicKey() == null || request.getEncryptedData() == null || request.getMac() == null) {
             return false;
         }
+        if (useTemporaryKeys == (request.getTemporaryKeyId() == null)) {
```
banterCZ, line 73: So you want to forbid the temporary key for other versions, right?

banterCZ (approved):
Great!

powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java
romanstrobl: Yes.

PR #713, opened 2025-01-30
Fix #712: Generify encryptor factory, encryptors and validators

### LINKED ISSUE #712: Generify encryptor factory, encryptors and validators
### Task specification

We should generify interfaces and classes for handling end-to-end encryption to allow multiple implementations.

romanstrobl:


jnpsk (approved):
looks ok

PR #717, opened 2025-02-07
Fix #711: Update hash-based counter implementation

### LINKED ISSUE #711: Update hash-based counter implementation
### Task specification

Hash-based counter should use KMAC-256.

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/generator/HashBasedCounter.java
```diff
@@ -31,34 +33,61 @@ public class HashBasedCounter implements Counter {
...
+     *
+     * @param version Protocol version.
+     * @throws GenericCryptoException Thrown in case version is not specified.
      */
-    public byte[] next(byte[] ctrData) {
-        byte[] nextCtrData = Hash.sha256(ctrData);
-        if (nextCtrData != null) {
-            return keyGenerator.convert32Bto16B(nextCtrData);
+    public HashBasedCounter(String version) throws GenericCryptoException {
+        if (version == null) {
```
petrdvorak, line 56: Here, I would maybe check if the version is a legitimate supported version.

petrdvorak (approved):
Looks OK, I added a minor comment about possible check (which is not necessary, the check seems to be done later in init method).

PR #719, opened 2025-02-11
Fix #710: Review usage of custom string in KDF

### LINKED ISSUE #710: Review usage of custom string in KDF
### Task specification

Currently, a common custom string is used for key derivation: `PA4KDF`. It would be useful to separate derived key spaces by using a different custom string for each derivation use case. Currently we can only add context when deriving keys which shifts the key derivation index, but keys derived this way can be used for different use cases without restriction.

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/kdf/Kdf.java
```diff
@@ -31,36 +31,33 @@
...
+     * @param context Key context allows derivation of multiple keys from the same source key material.
+     * @param index Optional byte array that can provide an additional key separation.
      * @param outLength Requested derived key size.
-     * @param context Optional context to use during key derivation.
-     * @return Derived secret key.
+     * @return Derived secret key length in bytes.
      * @throws GenericCryptoException Thrown in case of any cryptography error.
      */
-    public static SecretKey derive(SecretKey key, long index, int outLength, byte[] context) throws GenericCryptoException {
+    public static SecretKey derive(SecretKey key, String context, byte[] index, int outLength) throws GenericCryptoException {
```
jnpsk, line 49: IMHO we should clean the naming a bit so there isn't a clash with the [standard](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-108r1-upd1.pdf). It talks about `context`, which is passed to KMAC as data input and `label`, which is passed to KMAC as customization string.

This method builds `label` as `prefix || context` and then uses `index` as the data input.
Changing the naming would also help with the clarity of the code. For example, the method is called as
```
Kdf.derive(key=key, context=KEY_ENC_CUSTOM_STRING, index=keyContext, outLength=KEY_LENGTH);
```

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/kdf/Kmac.java
```diff
@@ -53,7 +98,6 @@ public static byte[] kmac256(SecretKey key, byte[] data, int outLength, byte[] c
             throw new GenericCryptoException("Invalid output length for KDF");
         }
         final KMAC kmac = new KMAC(KMAC_BIT_LENGTH, customString);
-        final byte[] keyBytes = key.getEncoded();
         if (keyBytes == null) {
             throw new GenericCryptoException("Secret key encoding is null");
         }
```
jnpsk, line 103: This is now redundant, it is validated already at the start of the method.

jnpsk (approved):
Looks ok

PR #723, opened 2025-02-27
Fix #706: Update data digest calculation

### LINKED ISSUE #706: Update data digest calculation
### Task specification

We should use `KMAC-256` for data digest calculation.

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/server/util/DataDigest.java
```diff
@@ -96,6 +99,24 @@ public DataDigest(int length) throws GenericCryptoException {
         }
         this.authorizationCodeLength = length;
     }
+    
+    /**
+     * Data digest of the list with string elements. Data is first normalized (items concatenated
+     * by '&amp;' character), then a random key is generated and hash (HMAC-SHA256) is computed. Finally,
+     * the resulting MAC is decimalized to the signature of a length 8 numeric digits.
+     * <p>
+     * In case the digest could not be computed, null value is returned.
```
jnpsk, line 108: Is it still true, as the validations now produce exceptions instead of returning null?

jnpsk (approved):
looks ok

PR #732, opened 2025-03-26
Fix #731: Use correct algorithm for EC key generation

### LINKED ISSUE #731: Use correct algorithm for EC key generation

_No response_

### Environment

_No response_

The following code is slightly incorrect:

```java
KeyPairGenerator.getInstance("ECDH", PowerAuthConfiguration.CRYPTO_PROVIDER_NAME);
```

The best practice is to use:
```java
KeyPairGenerator.getInstance("EC", PowerAuthConfiguration.CRYPTO_PROVIDER_NAME);
```

### Log output

_No response_

romanstrobl:


jnpsk (approved):
Good catch

PR #745, opened 2025-05-05
Fix #744: Update terminology for authentication codes

### LINKED ISSUE #744: Update terminology for authentication codes
### Task specification

Instead of using `symmetric signature`, use `authentication code`.

romanstrobl:
I updated terminology for authentication codes, so that the `signature` terminology is only used for asymmetric signatures.


jnpsk (approved):
Looks ok

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/authentication/AuthenticationKeyFactory.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * <p><b>PowerAuth protocol versions:</b>
+ * <ul>
+ *     <li>4.0</li>
+ * </ul>
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ *
+ */
+public class AuthenticationKeyFactory {
```
jnpsk, line 41: Will there be also `PowerAuthClientKeyFactory` alternative for v4 in the future? I am asking to know if we shouldn't keep the Server / Client separation in the class name.

powerauth-java-http/src/main/java/com/wultra/security/powerauth/http/PowerAuthHttpBody.java
```diff
@@ -28,14 +28,14 @@
 public class PowerAuthHttpBody {
 
     /**
-     * Prepare signature base string ("data to be signed") using request parameters.
+     * Prepare authentication base string ("data to be authenticated") using request parameters.
```
jnpsk, line 31: I am not sure if the "data to be authenticated" describes whats happening here. IMHO it could be deleted.

powerauth-java-http/src/main/java/com/wultra/security/powerauth/http/PowerAuthAuthorizationHttpHeader.java
```diff
@@ -101,23 +119,23 @@ public static class Key {
...
+    public PowerAuthAuthorizationHttpHeader() {
     }
 
     /**
      * Constructor with all required parameters.
      * @param activationId Activation ID.
      * @param applicationKey Application key.
-     * @param signature PowerAuth signature value.
-     * @param signatureType PowerAuth signature type.
+     * @param authCode PowerAuth authorization code value.
```
jnpsk, line 129: Is it `authorization code` or `authentication code`?

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/authentication/AuthenticationKeyFactory.java
```diff
@@ -0,0 +1,158 @@
...
+     * <ul>
+     *     <li>4.0</li>
+     * </ul>
+     *
+     * @param keyActivationSecret Activation shared secret key.
+     * @return An instance of factor key {@code KEY_AUTHENTICATION_CODE_POSSESSION}.
+     * @throws GenericCryptoException In case key derivation fails.
+     */
+    public SecretKey generatePossessionFactorKey(SecretKey keyActivationSecret) throws GenericCryptoException {
+        return KeyFactory.deriveKeyAuthenticationCodeKnowledge(keyActivationSecret);
```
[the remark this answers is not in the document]
romanstrobl: Good catch!

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/authentication/AuthenticationKeyFactory.java
romanstrobl: We do not need the server / client separation for authentication codes, the logic is the same. The separation is not too useful for most use cases, so I plan to keep and clean it up at the end of this project.

powerauth-java-http/src/main/java/com/wultra/security/powerauth/http/PowerAuthHttpBody.java
romanstrobl: I will update it to say "data used during authentication".

powerauth-java-http/src/main/java/com/wultra/security/powerauth/http/PowerAuthAuthorizationHttpHeader.java
romanstrobl: Should be `authentication code`, but note that the header is an `authorization` header (similar to HTTP basic), so the terminology clash may be present.

PR #746, opened 2025-05-15
Add test vector generator for Kdf and Aead algorithms

hvge:
This PR adds new routines that generate test vectors for Kdf and Aead routines.


powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -61,6 +64,7 @@ public class GenerateVectorDataTest {
     private static ObjectMapper objectMapper;
 
     private static final KeyConvertor KEY_CONVERTOR = new KeyConvertor();
+    private static final Random RANDOM = new Random();
```
romanstrobl, line 67: Can we use `SecureRandom`? `Random` in Java should not be used for any cryptography, even this is just a test.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+     * Generate random string or select random string from predefined set.
+     * @param minLength Minimum string length.
+     * @param maxLength Maximum string length.
+     * @param predefinedSet Predefined set of strings.
+     * @return Random string.
+     */
+    String getRandomString(int minLength, int maxLength, String[] predefinedSet) {
+        boolean fixedLength = minLength == maxLength;
+        int length = fixedLength ? maxLength : maxLength - minLength;
+        length = RANDOM.nextInt(length);
```
romanstrobl, line 112: Let's not reuse the variable `length`.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+        if (length == 0 && predefinedSet != null) {
+            return predefinedSet[RANDOM.nextInt(predefinedSet.length)];
+        }
+        if (!fixedLength) {
+            length += minLength;
+        }
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < length; i++) {
```
romanstrobl, line 120: If I use `minLength=5` and `maxLength=5`, the maximum random number for `lenght` is 4. Then we append 4 codepoints. I think you wanted to append 5 codepoints in this case, right?

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+    }
+
+    /**
+     * Generate random bytes with random length. The length is within specified range.
+     * @param minLength Minimum requested number of bytes.
+     * @param maxLength Maximum requested number of bytes.
+     * @return Random data.
+     */
+    byte[] getRandomBytes(int minLength, int maxLength) {
+        int size = minLength == maxLength ? minLength :  minLength + RANDOM.nextInt(maxLength - minLength);
```
romanstrobl, line 133: Extra space after `:`.

romanstrobl, line 133: I think there is the same off-by-1 problem as in the previous method, if I understand the intention correctly?

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+
+
+    /**
+     * Generate test data for common KDF function based on KMAC-256, used in protocol V4.
+     * @throws Exception In case any unknown error occurs.
+     */
+    @Test
+    void testKdfV4() throws Exception {
+        TestSet testSet = new TestSet("v4-kdf.json", "Test vectors for common KMAC-256 based KDF function used in protocol V4");
+        final String[] labels = {
```
romanstrobl, line 149: The labels are already predefined in class `KeyLabel`:

https://github.com/wultra/powerauth-crypto/blob/features/crypto4/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/kdf/KeyLabel.java

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+        writeTestVector(testSet);
+    }
+
+    /**
+     * Generate test data for KDF function for passwords, based on KMAC-256, used in protocol V4.
+     * @throws Exception In case any unknown error occurs.
+     */
+    @Test
+    void testPasswordKdfV4() throws Exception {
+        final String[] wellKnonwPassowrds = {
```
romanstrobl, line 183: Typo `wellKnonwPassowrds` should be `wellKnownPasswords`.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
```diff
@@ -85,6 +89,150 @@ public static void setUp() {
...
+            input.put("nonce", Base64.getEncoder().encodeToString(nonce));
+            input.put("associatedData", Base64.getEncoder().encodeToString(associatedData));
+            input.put("plaintext", Base64.getEncoder().encodeToString(plaintext));
+            final Map<String, String> output = new HashMap<>();
+            output.put("ciphertext", Base64.getEncoder().encodeToString(ciphertext));
+            testSet.addData(input, output);
+        }
+        writeTestVector(testSet);
+
+        writeTestVector(testSet);
```
romanstrobl, line 233: Why is `writeTestVector(testSet);` called twice?

romanstrobl (changes requested):
Let's cleanup the code a bit before merging.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
hvge: Yes, you're right. I would expect that some random data will be 32 bytes long if I choose `maxLength = 32`.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/util/GenerateVectorDataTest.java
hvge: You're right. I fixed this and also increased probability of selection from the predefined set.

romanstrobl (approved):
Perfect.

Issue #715, opened 2025-02-06
Test non-deterministic parts of V4 cryptography using mocking


romanstrobl:
### Description

Enrich the current `SharedSecret` test vectors with keys and other non-deterministic values used in ML-KEM and use the Mockito framework to inject such values for testing purposes. Similar approach can be used for deterministric end-to-end encryption using test vectors by injecting `nonce` a `timestamp` values during encryption.

PR #748, opened 2025-05-18
Fix #747: Add an option to generate test vectors for shared secret

### LINKED ISSUE #747: Add an option to generate test vectors for shared secret
### Task specification

It should be possible to generate test vectors from the tests.

romanstrobl:
Added automated generation of test vectors.


powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/lib/v4/sharedsecret/SharedSecretHybridTest.java
```diff
@@ -109,4 +109,31 @@ public void testEcdheMlkemWithTestVectors(Map<String, String> vector) throws Exc
...
+            ResponseCryptogram serverResponse = sharedSecretHybrid.generateResponseCryptogram(clientRequest);
+            SecretKey derivedSharedSecret = sharedSecretHybrid.computeSharedSecret(
+                    clientContext,
+                    (SharedSecretResponseHybrid) serverResponse.getSharedSecretResponse()
+            );
+            Map<String, String> vector = Map.of(
+                    "ecClientPrivateKey", Base64.getEncoder().encodeToString(KEY_CONVERTOR_EC.convertPrivateKeyToBytes(clientContext.getEcPrivateKey())),
+                    "pqcClientPrivateKey", Base64.getEncoder().encodeToString(KEY_CONVERTOR_PQC.convertPrivateKeyToBytes(clientContext.getPqcKemDecapsulationKey())),
+                    "ecServerPublicKey", ((SharedSecretResponseHybrid) serverResponse.getSharedSecretResponse()).getEcServerPublicKey(),
+                    "pqcEncapsulation", ((SharedSecretResponseHybrid) serverResponse.getSharedSecretResponse()).getPqcEncapsulation(),
```
hvge, line 129: The `pqcEncapsulation` is a bit misleading. It's `ciphertext` in NIST terminology or "wrapped key" (OpenSSL)

hvge (approved):
Looks OK. I would only consider rename `pqcEncsapsulation` property to something else.

romanstrobl:
Added deterministic test for low level KEM key generation, secret encryption and secret extraction. Both seed-only and full private keys are tested. I think we can close https://github.com/wultra/powerauth-crypto/issues/715 because instead of mocking I use the low level Bouncy Castle code.

PR #752, opened 2025-05-20
Fix #751: ML-KEM is not compatible with OpenSSL

### LINKED ISSUE #751: ML-KEM is not compatible with OpenSSL

Branch `features/crypto4`.

### Environment

None.

Update ML-KEM for compatibility with OpenSSL. The test vectors do not match.

romanstrobl:


hvge (approved):
Looks OK. I've tested this change against my OpenSSL wrapper and we can finally exchange keys with using ML-KEM.

PR #757, opened 2025-05-26
Fix #756: Separate v4 server and client classes

### LINKED ISSUE #756: Separate v4 server and client classes
### Task specification

Currently the server and client classes are used both for v3 and v4 protocols.

We should separate the code to avoid mixing various v3 and v4 methods in each of the classes and allow future deprecation of v3 code easily.

romanstrobl:
This pull requests separates v3 and v4 classes for client and server usage and cleans up the API to be used from client and server.

### DIFF
--- a/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/server/v4/token/ServerTokenGenerator.java
+++ b/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/server/v4/token/ServerTokenGenerator.java
@@ -14,24 +14,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.wultra.security.powerauth.crypto.server.token;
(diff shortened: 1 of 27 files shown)


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/server/v4/token/ServerTokenGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Generate random token ID. Use UUID format.
+     * @return Random token ID.
+     */
+    public String generateTokenId() {
+        return tokenUtils.generateTokenId();
+    }
+
+    /**
+     * Generate random token secret, 16 or 32 random bytes depending on the version.
```
jnpsk, line 47:
```suggestion
     * Generate random 32 byte long token secret.
```

jnpsk (approved):
Looks good.
Maybe some of the version validations are redundant now.

PR #763, opened 2025-06-11
Update shared secret test

hvge:
Added a test case that generates counterpart SharedSecretHybrid responses to requests created by the mobile client. This tests the full request-response cycle using data originating from both parties.


romanstrobl (approved):
Thanks for the test vectors.

PR #767, opened 2025-06-13
Fix #766: Generate test vectors for JWT used in temporary keys

### LINKED ISSUE #766: Generate test vectors for JWT used in temporary keys
### Task specification

Generate test vectors for JWT requests and responses used in temporary keys and validating signatures with different algorithms.

romanstrobl:
Test vectors for JWT requests and responses for algorithms `EC_384` and `EC_384_ML_L3`.


hvge (approved):
Looks OK

PR #778, opened 2025-07-03
Fix #768: Improve crypto agility for PQC KEM and DSA

### LINKED ISSUE #768: Improve crypto agility for PQC KEM and DSA
### Description

Improve crypto agility by switching from concrete classes to interfaces for:
- PQC KEM
- PQC DSA
- associated key convertors

Allow instantiation based on PQC algorithms with generic interfaces and swappable implementations.

### Acceptance criteria

It is possible to swith different PQC algorithms easily.

### Technical specification

Follow PQC algorithm names based on standards.

### QA specification

Tested by unit tests.

romanstrobl:
I switched to interfaces so we can change algorithms easily in the future.


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/PqcDsaKeyConvertor.java
```diff
@@ -1,19 +1,3 @@
-/*
```
jnpsk: In several files there is missing license or has been removed.

jnpsk (approved):
Looks good

PR #780, opened 2025-07-03
Fix #779: Implement runtime key protection

### LINKED ISSUE #779: Implement runtime key protection
### Description

In some rare situations we must reveal the sensitive keys into the JVM runtimes. The best example is vault-unlock operation when the server returns sensitive key derivate encoded in JSON.

To obfuscate such data in the memory we use simple encryption scheme based on AES.

### Acceptance criteria

Sensitive keys are protected using runtime protection.

### Technical specification

- `byte[] RP.wrapKey(SecretKey key, String keyContext)`
  - Function obfuscate sensitive key derived from `KEY_ACTIVATION_SECRET`. Parameters:
    - `key` - key to obfuscate
    - `keyContext` - key context, see `RP_` constants in [Index registry](crypto4-keys.md#index-registry)
  - Implementation:
    ```java
    byte[] KEY_BYTES = KeyConversion.getBytes(key);
    byte[] SECRET = KDF.derive(KDK_UTILITY, keyContext, null, 32 + 16);  // 32 bytes key, 16 bytes IV
    SecretKey kek = KeyConversion.secretKeyFromBytes(ByteUtils.subarray(SECRET, 0, 32));
    byte[] iv = ByteUtils.subarray(SECRET, 32, 16);
    // It's OK to use the same IV for multiple times, because we encrypt the same key
    // with the same KEK. This limits this scheme to keys derived from the common
    // shared secred.
    byte[] C_KEY = AES.encrypt(KEY_BYTES, iv, kek, "AES/CTR/NoPadding");
    return C_KEY;
    ```

- `SecretKey RP.unwrapKey(byte[] wrappedKey, long index)`
  - Function extract sensitive key previously obfuscated by `RP.wrapKey()` function. Parameters:
    - `wrappedKey` - obfuscated key
    - `keyContext` - key context, see `RP_` constants in [Index registry](crypto4-keys.md#index-registry)
  - Implementation:
    ```java
    byte[] SECRET = KDF.derive(KDK_UTILITY, keyContext, null, 32 + 16);  // 32 bytes key, 16 bytes IV
    SecretKey kek = KeyConversion.secretKeyFromBytes(ByteUtils.subarray(SECRET, 0, 32));
    byte[] iv = ByteUtils.subarray(SECRET, 32, 16);
    // Decrypt wrapped key
    byte[] KEY_BYTES = AES.decrypt(wrappedKey, iv, kek, "AES/CTR/NoPadding");
    return KeyConversion.secretKeyFromBytes(KEY_BYTES);
    ```

The scheme above can be applied only to the keys that are derived from the shared secret (e.g. `KEY_ACTIVATION_SECRET`). The reason for this is that we reuse IV with the same encryption key.

All wrapping keys with `RP_` prefix are constructed in this way.

### QA specification

Tested by end-to-end tests.

romanstrobl:
Key protection for transport within unprotected runtimes.

### DIFF
--- a/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/kdf/KeyFactory.java
+++ b/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/kdf/KeyFactory.java
@@ -220,6 +220,16 @@ public class KeyFactory {
     }
 
     /**
+     * Derives {@code KDK_UTILITY} from {@code KEY_ACTIVATION_SECRET}.
+     * @param keyActivationSecret The activation secret key.
+     * @return Derived KDK for utility purposes.
+     * @throws GenericCryptoException In case of cryptographic failure.
+     */
+    public static SecretKey deriveKdkUtility(SecretKey keyActivationSecret) throws GenericCryptoException {
+        return derive(keyActivationSecret, KeyLabel.UTIL);
+    }
+
+    /**
      * Derives {@code KDK_AUTHENTICATION_CODE} from {@code KEY_ACTIVATION_SECRET}.
      * @param keyActivationSecret The activation secret key.
      * @return Derived KDK for authentication codes.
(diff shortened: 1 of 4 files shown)


romanstrobl:
This functionality will not be needed, closing pull request.

PR #787, opened 2025-08-04
Fix #704: Update activation methods for crypto4

### LINKED ISSUE #704: Update activation methods for crypto4
### Description

The changes in cryptography should be reflected in activation support classes.

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

_No response_

romanstrobl:
I implemented v4 activation fingerprints and activation code signatures.

I also updated tests to cover the v4 scenarios and updated test vectors.

I plan to add more tests in future pull requests to cover all v4 changes.

### DIFF
--- a/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/ECPublicKeyFingerprint.java
+++ b/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/ECPublicKeyFingerprint.java
@@ -29,6 +29,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.interfaces.ECPublicKey;
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * Class that is used for computing EC public key fingerprint. Goal of the public key fingerprint is to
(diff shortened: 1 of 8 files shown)


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/ECPublicKeyFingerprint.java
```diff
@@ -57,36 +58,35 @@ public static String compute(ECPublicKey devicePublicKey, ECPublicKey serverPubl
...
-                    byte[] devicePublicKeyBytes = toByteArray(devicePublicKey);
-                    byte[] activationIdBytes = activationId.getBytes(StandardCharsets.UTF_8);
-                    byte[] serverPublicKeyBytes = toByteArray(serverPublicKey);
-                    ByteBuffer dataBuffer = ByteBuffer.allocate(devicePublicKeyBytes.length + activationIdBytes.length + serverPublicKeyBytes.length);
-                    dataBuffer.put(devicePublicKeyBytes);
-                    dataBuffer.put(activationIdBytes);
-                    dataBuffer.put(serverPublicKeyBytes);
-                    fingerprintData = dataBuffer.array();
+            final byte[] fingerprintData;
+            if (Objects.requireNonNull(activationVersion) == ActivationVersion.VERSION_3) {
```
jnpsk: This will throw NPE if the `activationVersion` is null. Wouldn't be more convenient to just jump into the else branch if the `activationVersion` is null? Ideally, you can just check if the version equals v3 before the try block, similarly as for the v4 variant.

jnpsk (approved):
looks good

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/ECPublicKeyFingerprint.java
romanstrobl: Ok, I will change it, too.

PR #793, opened 2025-08-08
Fix #792: Update authentication code component size for crypto4

### LINKED ISSUE #792: Update authentication code component size for crypto4
### Task specification

The authentication code component size should be 32 bytes for crypto4.

romanstrobl:
Updated authentication code component size to 32B for crypto4.


jnpsk (approved):
looks ok

PR #796, opened 2025-08-13
Fix #795: Add interfaces for PQC convertors

### LINKED ISSUE #795: Add interfaces for PQC convertors
### Task specification

Introduce interfaces for PQC convertors to allow crypto agility.

romanstrobl:
I switched to interfaces for convertors and cleaned up the PQC implementations.


jnpsk (approved):
looks good

PR #800, opened 2025-09-05
Fix #799: ByteUtils.subarray should accept 0 length

### LINKED ISSUE #799: ByteUtils.subarray should accept 0 length

develop

### Environment

local

Currently the method ByteUtils.subarray() does not accept zero length. This can be a valid use case, so it should just return an empty array.

### Log output

_No response_

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/ByteUtils.java
```diff
@@ -138,9 +138,12 @@ public static byte[] subarray(byte[] array, int start, int length) {
         if (start < 0) {
             throw new IllegalArgumentException("Invalid start index");
         }
-        if (length <= 0) {
+        if (length < 0) {
             throw new IllegalArgumentException("Invalid length");
         }
+        if (length == 0) {
+            return new byte[0];
+        }
```
jnpsk: Since the zero-length check runs before validating array bounds, the method accepts calls like `ByteUtils.subarray(new byte[]{1, 2, 3}, 50, 0)`.
I’d either move that explicit check after the index validation or remove it entirely and let `System#arraycopy` handle the zero-length case too.

romanstrobl: Good point, I will move it after.

jnpsk (approved):
looks ok

PR #814, opened 2025-09-22
Fix #813: Update tests for BouncyCastle 1.82

### LINKED ISSUE #813: Update tests for BouncyCastle 1.82
### Task specification

Bouncy Castle changed the internal implementation in version `1.82` and the deterministic tests fail. We should either use the public API instead with a fixed random source (preferred) or migrate to the updated implementation.

romanstrobl:


banterCZ (approved):
Great! Without reflection, it should be much more stable.

PR #821, opened 2025-10-02
Fix #820: Use configured provider instead of "BC"

### LINKED ISSUE #820: Use configured provider instead of "BC"

develop

### Environment

_No response_

Instead of using "BC" in parameters, use the constant provided in the library.

### Log output

_No response_

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/ml/MlDsa.java
```diff
@@ -58,7 +59,7 @@ public MlDsa(MLDSAParameterSpec dsaParameterSpec) throws GenericCryptoException
     @Override
     public KeyPair generateKeyPair() throws CryptoProviderException {
         try {
-            final KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("MLDSA", "BC");
+            final KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("MLDSA", PowerAuthConfiguration.CRYPTO_PROVIDER_NAME);
```
banterCZ, line 62: Isn't the value also `BC`?

romanstrobl: Yes it is.

romanstrobl: The point is to avoid the same "BC" string in the codebase.

jnpsk (approved):

PR #824, opened 2025-10-29
Fix #823: Update key label and fingerprint computation for EC_P384_ML_L5

### LINKED ISSUE #823: Update key label and fingerprint computation for EC_P384_ML_L5
### Task specification

The algorithm `EC_P384_ML_L5` should use a specific key label and fingerprint.

romanstrobl:


jnpsk (approved):
Looks good

PR #841, opened 2025-11-21
Fix #838: Update public key validation during import

### LINKED ISSUE #838: Update public key validation during import
### Task specification

The subgroup check during imported public key validation is not required for NIST curves, because they use cofactor = 1. The check is only relevant for curves with cofactor > 1.

romanstrobl:
This change will improve performance because the validation can be simplified.

### DIFF
--- a/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/PublicKeyValidator.java
+++ b/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/PublicKeyValidator.java
@@ -31,7 +31,7 @@ public class PublicKeyValidator {
      * Validate the public key:
      * <ul>
      *   <li>check that the EC point is not the point at infinity</li>
-     *   <li>check that point order matches the order defined in EC curve</li>
+     *   <li>check that the curve cofactor is equal to 1 and subgroup order check can be omitted</li>
      * </ul>
      *
      * @param curve EC curve.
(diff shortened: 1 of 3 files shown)


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/PublicKeyValidator.java
```diff
@@ -44,10 +43,10 @@ public void validate(ECCurve curve, ECPoint point) throws GenericCryptoException
             throw new GenericCryptoException("Invalid public key with point equal to the point at infinity");
         }
 
-        final BigInteger n = curve.getOrder();
-        final ECPoint calculatedPoint = ECAlgorithms.referenceMultiply(point, n);
-        if (!calculatedPoint.isInfinity()) {
-            throw new GenericCryptoException("Point order does not match the order defined in EC curve");
+        // Subgroup order validation is not necessary for NIST curves because they have cofactor = 1.
+        // Check that cofactor = 1 to guard against accidentally supporting curves with cofactor > 1.
```
jnpsk, line 47: The method's JavaDoc should be changed too.

powerauth-java-crypto/src/test/java/com/wultra/security/powerauth/crypto/ec/PublicKeyValidatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Test for public key validation.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class PublicKeyValidatorTest {
```
jnpsk: Nitpick - public does not have to be present for test class and methods.

jnpsk (approved):
looks ok

PR #843, opened 2025-11-28
Fix #842: Fix issues found by coverity scan

### LINKED ISSUE #842: Fix issues found by coverity scan
### Task specification

Coverity scan found several unlogged exceptions.

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/dh/DhKem.java
```diff
@@ -116,6 +121,7 @@ public PublicKey convertBytesToPublicKey(byte[] publicKeyBytes) throws GenericCr
         try {
             return KEY_CONVERTOR.convertBytesToPublicKey(EcCurve.P384, publicKeyBytes);
         } catch (CryptoProviderException | InvalidKeySpecException e) {
+            logger.debug(e.getMessage(), e);
```
banterCZ, line 124: I think that Coverity minds that the security exception `InvalidKeySpecException` may be swallowed. It is passed as a cause of GenericCryptoException, and static analysis may lose the context. Anyway, I am ok with this solution, since `GenericCryptoException` is a checked exception.

banterCZ (approved):

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/dh/DhKem.java
romanstrobl: Yes, this is why I put it into debug.

PR #858, opened 2025-12-17
Fix #857: HPKE object is shared

### LINKED ISSUE #857: HPKE object is shared

_No response_

### Environment

_No response_

_No response_

### Log output

_No response_

jnpsk:


banterCZ (approved):
Good, not globally share variable. Just wondering whether `ThreadLocal` would fix it as well.

banterCZ (approved):
Yes, I like the method more.

Issue #867, opened 2026-01-13
Review and update documentation for cryptography


romanstrobl:
### Task specification

Cryptographic algorithms have been updated with crypto4, review and update all documentation including:
- updated crypto primitives
- migration to AEAD from ECIES
- crypto agility
- activation confirmation
- authentication code calculation
- upgrade process
- key derivation
- token calculation and validation
- dynamic factor keys
- vault unlock
- activation status
- shared secret derivation
- temporary key retrieval, signing and signature verification

Also update key labels such as `KEY_SERVER_MASTER_PRIVATE` and `KEY_SERVER_MASTER_PUBLIC`, they use specific algorithms now.

PR #876, opened 2026-01-27
Fix #875: Update basic cryptography documentation

### LINKED ISSUE #875: Update basic cryptography documentation
### Task specification

Update documents:
- Basic definitions
- Key derivation
- List of used keys
- Implementation notes

romanstrobl:


docs/Basic-definitions.md
```diff
@@ -8,60 +8,70 @@ The following basic cryptography algorithms and parameters are used in the Power
...
 byte[] original = AES.decrypt(byte[] encrypted, byte[] iv, SecretKey key, String transformation);
 ```
 
-### PBKDF2
+### Generic KDF
+
+Keys are derived from an original secret using hierarchical string labels to guarantee that derived keys are never reused for different purposes.
 
-An algorithm for key stretching, converts a short password into long key by performing repeated hash iteration on the original data. HMAC-SHA1 algorithm is used for a pseudo-random function. Implementations must make sure resulting key is converted into a format usable by the AES algorithm.
+Keys are never reused for multiple purposes; each derived key has a unique label.
```
jnpsk: The second paragraph seems redundant

docs/Basic-definitions.md
```diff
@@ -8,60 +8,70 @@ The following basic cryptography algorithms and parameters are used in the Power
...
 
 ```java
-byte[] bytes = KDF_X9_63_SHA256.derive(byte[] secret, byte[] info, int length);
+SecretKey expandedKey = KDF.derivePassword(byte[] password, byte[] salt);
 ```
 
 ### ECDSA Signatures
 
-An algorithm for elliptic curve based signatures, uses SHA256 hash algorithm and P256r1 EC curve. It defines the following operations:
+An algorithm for elliptic curve based signatures, uses SHA-384 hash algorithm and P-384 EC curve. It defines the following operations:
```
jnpsk: a nitpick
```suggestion
An algorithm for elliptic curve based signatures, uses SHA-384 hash algorithm and P-384 elliptic curve. It defines the following operations:
```

docs/Basic-definitions.md
```diff
@@ -79,60 +89,159 @@ Verify the signature for given data using a given public key.
...
 ```
 
-### ECDH Key Agreement
+### ML-DSA Signatures
 
-An algorithm for elliptic curve Diffie-Hellman, uses P256r1 curve. We define a single operation on ECDH, a symmetric key deduction between parties A and B:
+Post-quantum signature algorithm MLDSA can be used for post-quantum signatures.
 
-Derive a shared secret using a private key of party A and a public key of party B using the follwing method.
+```java
```
jnpsk, line 95: Maybe it's worth to describe those code snippets a bit too

docs/Implementation-notes.md
```diff
@@ -121,31 +148,33 @@ You can also check [Activation Code](./Activation-Code.md) document to get a mor
 
 ## Generating Key Pairs
 
-The device and server keys are generated using ECDH algorithm with P256 curve:
+The device and server ECC keys are generated using ECDSA / ECDH with **P-384** curve:
```
jnpsk: keys generated using ECDSA / ECDH?

docs/Key-derivation.md
```diff
@@ -1,98 +1,134 @@
...
+``` java
+SecretKey KEY_AUTHENTICATION_CODE_KNOWLEDGE = KDF.derive(KDK_AUTHENTICATION_CODE, "auth/knowledge");
 ```
 
-The key `KEY_SIGNATURE_KNOWLEDGE` is then decrypted using the inverse algorithm - the stored salt end entered password is used to decrypt the encrypted `C_KEY_SIGNATURE_KNOWLEDGE`.
+This key is stored encrypted using a key derived from a password or PIN code using the Password-based KDF.
+
+The knowledge factor key may change over time (for example after password change). The server maintains current and next versions of this key.
+
+Because unauthenticated encryption is used, decryption with an incorrect PIN produces random key material. This is intentional: verification always happens on the server side.
```
jnpsk, line 51: The former info box feels more explanatory.

docs/List-of-Used-Keys.md
```diff
@@ -2,30 +2,96 @@
...
+| name                                 | created as                                     | purpose                                                                                                                                                      |
+|--------------------------------------|------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| _key_master *_                       |                                                |                                                                                                                                                              |
+| `KEY_MASTER_P256_PRIVATE`            | ECDSA/ECDH - private key                       | Stored on server, used for backward compatibility with older clients                                                                                         |
```
jnpsk: In the E2EE or REST API doc chapters, we call keys differently than in this document, for example `KEY_SERVER_MASTER_PRIVATE` in REST API, but here it is `KEY_MASTER_{algID)_PRIVATE`

docs/List-of-Used-Keys.md
```diff
@@ -2,30 +2,96 @@
...
+| `KEY_MASTER_P256_PUBLIC`             | ECDSA/ECDH - public key                        | Used by older clients                                                                                                                                        |
```
jnpsk: "older clients" sounds vague.

docs/List-of-Used-Keys.md
```diff
@@ -2,30 +2,96 @@
 
 The following keys are used in the PowerAuth cryptography scheme.
 
-## Application Scoped Keys
-
-| name                        | created as                    | purpose                                                                                                                                                             |
-|-----------------------------|-------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `KEY_SERVER_MASTER_PRIVATE` | ECDH - private key            | Embedded on server, used to assure authenticity of data during the transfer from server to client during application scoped use-cases (i.e., device activation).    |
```
jnpsk, line 9: The former description feels more informative, I suggest to keep those.

docs/List-of-Used-Keys.md
```diff
@@ -2,30 +2,96 @@
...
+| _Other_                         |                                    |                                          |                                                                                                                            |
+| `other/expand-biometry-key`     | KEK_AUTHENTICATION_CODE_BIOMETRY   | 128-bit KEK_AUTHENTICATION_CODE_BIOMETRY | Expands 128 bit biometry factor related KEK to 256 bit                                                                     |
+
+### Notes
+
+> Note 1: In E2EE, `BASE_KEY` is equal to `KEY_TEMPORARY_SHARED_SECRET`.
+
+> Note 2: `DEVICE_KEY` is calculated as `Hash.sha3_256("device-specific-data")`
+
+> Note 3: `KEY_SHARED_SECRET` is KEY_ACTIVATION_SECRET or KEY_TEMPORARY_SECRET, depending on purpose of established secret.
```
jnpsk: What are those mentioned purposes?
Overall those notes seem too brief

docs/Basic-definitions.md
romanstrobl: Ah yes, i will remove it.

docs/Key-derivation.md
romanstrobl: The algorithm for encrypting the knowledge factor key is being reconsidered, so I intentionally leave the details out. We can add them later after the final algortihm is decided.

docs/List-of-Used-Keys.md
romanstrobl: Ok, I will resolve this issue separately after merging all pull requests in the review ticket.

romanstrobl: Added: https://github.com/wultra/powerauth-crypto/issues/867

docs/List-of-Used-Keys.md
romanstrobl: This is explained in other chapters - Key-derivation and Implementation-notes.

docs/List-of-Used-Keys.md
romanstrobl: Good idea.

jnpsk (approved):

PR #889, opened 2026-02-20
Revise algorithm suite IDs descriptions in docs

petrdvorak:
Updated terminology and descriptions for cryptographic algorithm suite IDs in the documentation.


romanstrobl (approved):
Nice improvements!

PR #895, opened 2026-03-17
Fix #894: Allow validation of encrypted requests without body

### LINKED ISSUE #894: Allow validation of encrypted requests without request data
### Task specification

For the use case of validating data for obtaining an encryptor, the validation of encrypted data in requests and mac should be optional.

romanstrobl:


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/encryptor/aead/AeadRequestResponseValidator.java
```diff
@@ -50,11 +50,11 @@ public AeadRequestResponseValidator(String protocolVersion) throws AeadException
     }
 
     @Override
-    public boolean validateEncryptedRequest(AeadEncryptedRequest request) {
+    public boolean validateEncryptedRequest(AeadEncryptedRequest request, boolean validateRequestData) {
```
jnpsk: I am not a fan of adding a boolean flag to the validation method, especially when it is to handle "overloaded meaning" of `AeadEncryptedRequest` in single specific scenario on PAS and thus _almost_ every time set to true (if I am not overlooking something).

I would consider:
- Just using empty byte array instead of null in the `AeadEncryptedRequest.encryptedData` in the PAS specific scenario.
- Postpone the `encryptedData` validation to the time just before the actual data encryption is going to happen, as apparently there is a scenario where null `encryptedData` is also a valid case.
- Create a new method `validateEncryptedRequestWithoutData()` or something like that to avoid passing a flag every time, and keep the current as default.

jnpsk (commented):
This looks functional overall.
I’d like to share a suggestion that, from my perspective, could help reduce the code complexity.

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/v4/encryptor/aead/AeadRequestResponseValidator.java
romanstrobl: Note that this is not only about the encypted data, but also about the data mac.

We can handle the validation using a separate method. The use cases for obtaining the encryptor and using the encryptors differ. Thanks for the feedback.

jnpsk (approved):
Looks good,
it may be further optimized by using `validateEncryptedRequestWithoutData` from within the `validateEncryptedRequest` method to avoid code duplicity.

romanstrobl:
Ok I will remove the duplication.

jnpsk (approved):
👍

PR #898, opened 2026-03-23
Fix #896: Move SdkConfigurationSerializer

### LINKED ISSUE #896: Move SdkConfigurationSerializer
### Task specification

Currently the serializer is duplicated in powerauth-server and powerauth-cmd-tool.

jnpsk:
Move the `SdkConfigurationSerializer` functionality from PAS.
- The DataWriter/Reader interfaces were dropped as they were no use for it, instead implementation is used directly.
- Introducing dedicated `SdkConfigurationException` to handle a specific errors from the de/serialization
- The core functionality was copy-pasted


romanstrobl (approved):
Looks OK.

PR #899, opened 2026-03-30
Fix #837: Remove classloader workaround in public and private key conversion

### LINKED ISSUE #837: Remove classloader workaround in public and private key conversion
### Task specification

The public and private key conversion can be simplified, the workaround for classloader issues is no longer needed. Currently the library is used with one BC loaded using Security.addProvider() because ML-KEM suffers from similar classloader issues as handled by the workaround in class `KeyConvertor`.

romanstrobl:
There was a workaround for classloading issues with BC which made the key conversion more complex. This is no longer required, because it is guaranteed that BC library is loaded on once on the server. 

I switched to using BC instead of Java Security API to make the conversion more straightforward and added missing checks.

### DIFF
--- a/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/KeyConvertor.java
+++ b/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/KeyConvertor.java
@@ -22,6 +22,9 @@ import com.wultra.security.powerauth.crypto.lib.model.exception.CryptoProviderEx
 import com.wultra.security.powerauth.crypto.lib.model.exception.GenericCryptoException;
 import org.bouncycastle.jce.ECNamedCurveTable;
 import org.bouncycastle.jce.spec.ECNamedCurveParameterSpec;
+import org.bouncycastle.jce.spec.ECPrivateKeySpec;
+import org.bouncycastle.jce.spec.ECPublicKeySpec;
+import org.bouncycastle.math.ec.ECPoint;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -31,7 +34,7 @@ import java.math.BigInteger;
 import java.security.*;
 import java.security.interfaces.ECPrivateKey;
 import java.security.interfaces.ECPublicKey;
-import java.security.spec.*;
+import java.security.spec.InvalidKeySpecException;
 
 /**
  * Key convertor for conversion of symmetric and asymmetric keys.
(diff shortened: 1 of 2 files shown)


powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -65,16 +69,24 @@ public byte[] convertPublicKeyToBytes(PublicKey publicKey) throws CryptoProvider
...
+        final java.security.spec.ECPoint ecPoint = ecPublicKey.getW();
         // Create EC point using Bouncy Castle library
-        ECNamedCurveParameterSpec ecSpec = ECNamedCurveTable.getParameterSpec(curve.getName());
+        final ECNamedCurveParameterSpec ecSpec = ECNamedCurveTable.getParameterSpec(curve.getName());
         if (ecSpec == null) { // can happen with incorrectly initialized crypto provider.
-            throw new CryptoProviderException("Crypto provider does not support curve " + curve.getName());
+            throw new CryptoProviderException("Crypto provider does not support EC curve " + curve.getName());
+        }
+        if (!(ecPublicKey.getParams().getOrder().equals(ecSpec.getN()))) {
+            throw new GenericCryptoException("Public key EC curve order does not match EC curve order");
```
jnpsk: nitpick - the message seems a bit confusing. Maybe just unexpected EC curve order, or does not match expected EC curve order

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/KeyConvertor.java
```diff
@@ -65,16 +69,24 @@ public byte[] convertPublicKeyToBytes(PublicKey publicKey) throws CryptoProvider
      * @param publicKey An EC public key to be converted.
      * @return A byte array representation of the EC public key.
      * @throws CryptoProviderException When crypto provider is incorrectly initialized.
+     * @throws GenericCryptoException When public key is invalid.
      */
-    public byte[] convertPublicKeyToBytes(EcCurve curve, PublicKey publicKey) throws CryptoProviderException {
+    public byte[] convertPublicKeyToBytes(EcCurve curve, PublicKey publicKey) throws CryptoProviderException, GenericCryptoException {
```
jnpsk, line 74: Isn’t this method used by other components? Introducing an additional exception may require corresponding changes in the callers.

jnpsk (approved):
OK!

powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/util/KeyConvertor.java
romanstrobl: Yes, I will also handle this in other components.

PR #916, opened 2026-04-17
Fix #907: Validate status blob size

### LINKED ISSUE #907: Validate status blob size
### Task specification

The status blob size is not validated in binary status blob.

jnpsk:
Allow appended MAC for V4


banterCZ (approved):
Thanks! Let's test that.

PR #933, opened 2026-05-04
Fix #932: Update cryptography documentation

### LINKED ISSUE #932: Update cryptography documentation
### Task specification

Add or update chapters of documentation for following topics:
- SDK configuration binary blob
- update activation fingerprints and signatures
- KMAC-256 key derivation customization strings
- replay attack prevention

romanstrobl:


docs/Key-derivation.md
```diff
@@ -120,7 +129,7 @@ SecretKey KDK_VAULT = KDF.derive(KEY_ACTIVATION_SECRET, "vault");
 SecretKey KEK_DEVICE_PRIVATE = KDF.derive(KDK_VAULT, "vault/kek-device-private");
 ```
 
-The original device private key is stored encrypted using AEAD. The vault encryption key is never stored directly. It is always re‑derived from `KEY_ACTIVATION_SECRET` when needed.
+The original device private key is stored encrypted using AES-256-GCM. The vault encryption key is never stored directly. It is always re‑derived from `KEY_ACTIVATION_SECRET` when needed.
```
hvge: Is this the situation on the server or on the client? If it's the client, then we're really using AEAD for the encryption.

hvge (commented):
This is a nice improvement. I have found just one questionable change. See below.

docs/Key-derivation.md
hvge: If this is a mobile, then the mobile client doesn't know `KEY_ACTIVATION_SECRET` after the activation is complete. We no longer re-derivate the keys once the activation is complete. In V4, the private key is unlocked only for the purpose of digital signature calculation.

romanstrobl: I removed this section, it is unclear.

hvge (approved):
👍

PR #938, opened 2026-05-18
fix #936: register lombok annotation processor

### LINKED ISSUE #936: Fix: Support Lombok with Java 25
### Task specification

Lombok 1.18.x relies on implicit annotation processor discovery which is broken on Java 25 — the `logger` field (and other generated members) are not visible to `javac`:

```
cannot find symbol
  symbol: variable logger
```

Affects local development on Java 25 (any vendor). CI uses Java 21 and is unaffected.

**Root cause:** Starting with JDK 23, the default annotation processing policy changed to `none`.

**Fix:** Explicitly register Lombok via `annotationProcessorPaths` in the root `pom.xml`'s `maven-compiler-plugin` configuration.

See: wultra/tasklist#898 for parent epic.

vita-kotacka:
## Description

Registered Lombok as an explicit annotation processor in the root Maven compiler plugin configuration. The existing module Lombok dependency now reuses the parent `lombok.version` property.

## Motivation and Context

Java 25 no longer enables implicit annotation processor discovery by default, so Lombok-generated members may be unavailable during compilation. Explicitly configuring Lombok in `annotationProcessorPaths` keeps local Java 25 builds working.

## Closes

Closes #936

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,7 @@
         <maven-javadoc-plugin.version>3.12.0</maven-javadoc-plugin.version>
         <maven-source-plugin.version>3.4.0</maven-source-plugin.version>
         <maven-surefire-plugin.version>3.5.5</maven-surefire-plugin.version>
+        <lombok.version>1.18.46</lombok.version>
         <slf4j.version>2.0.18</slf4j.version>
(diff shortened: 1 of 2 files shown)


romanstrobl (approved):
Looks OK.

